PR #7, opened 2023-03-07
ThePortugueseMan.CodingTracker

ThePortugueseMan:


ThePortugueseMan:
Codacy says I dont't need System.Configuration to access the COnfigurationManager which is not true, that's why the check failed

TheCSharpAcademy:
> Codacy says I dont't need System.Configuration to access the ConfigurationManager which is not true, that's why the check failed

True! Codacy does that sometimes 😟

TheCSharpAcademy:
@ThePortugueseMan Hey there! Well done! Very impressed with the project 😃

I'm approving it now but I'd just like to comment that ideally you should tell the user what's the minimum time that session should have to be registered: 

image

ThePortugueseMan:
Thanks for the kind words! 😃

I added that information in a new commit!

Thanks for the review!

PR #11, opened 2023-04-20
Coding Tracker project

LucianoNicolasArrieta:


TheCSharpAcademy (changes requested):
Hey there! The project is looking good but I'm getting this error: 

image

LucianoNicolasArrieta:
Hi! fixed. I don't really know what happened because it works with some dates but with others not, i just split the method that validates the end date into two steps and it seems now it's working.

TheCSharpAcademy:
@LucianoNicolasArrieta hey there! I'm approving since all requirements are met but I'm not being able to see the reports: 

image

Also the app breaks if I don't have any records: 

image

PR #12, opened 2023-04-27
Project Finished

Chad1082:


Chad1082:
Hi, this failing on unnecessary usings, but if I remove the using the project wont compile as it's needed for the Configuration Manager.

TheCSharpAcademy:
@Chad1082 I'm approving the project. Just for future reference, with using statements, conn.Close() isn't necessary as the the using manages the disposal of the connection: 

image

Also be aware of unnecessary empty lines (127, 128):
image

PR #13, opened 2023-05-04
Program to track how long you code. Can see total time coded. Users c…

heartingNinja:
…an set time manually or using system time. I am not sure if I set up the classes in the correct way.


TheCSharpAcademy:
@heartingNinja brilliant work! very impressed. I particularly like the possibility to add a record using the system's time 😁

PR #19, opened 2023-06-06
Coding tracker review request

ohshie:
Hey, im back. Finally somewhat finished with a lot of irl stuff. 
Anyhow. All basic requirements from project is done. Somewhere in the middle of writing third menu i was so done with my way of doing it that i decided to look for a better console output and found pretty kewl lib called SpectreConsole that i decided to use. 
I still kept verify class/methods with my own implementation (although im not quite sure they are correct!), but spectreconsole actually gives a nice verification on its own.


Ohshie.CodingTracker/Ohshie.CodingTracker/SessionsOperator/NewCodingSession.cs
```diff
@@ -0,0 +1,83 @@
...
+            Description = sessionDescription,
+            Date = _time
+        };
+        
+        _stopwatch.Start();
+        ConsoleProgress();
+        _stopwatch.Stop();
+
+        newSession.Length = _stopwatch.Elapsed.ToString(@"HH\:mm\:ss");
+
```
TheCSharpAcademy, line 56: I'm getting this error when stopping the stop watch, could you please check? 😊
image

TheCSharpAcademy (changes requested):

TheCSharpAcademy:
@ohshie Good work! I did find the "Verify" name not very useful. What is it verifying? I think "Validation" would have been better. 😊 Approved anyways

PR #25, opened 2023-07-28
Project Finished

CarDioLogic:
Here it is. Project Finished! :)


CarDioLogic:
Codady keeps "flagging" to delete the following line:
CodingTrackerCarDioLogics/CodingTrackerJMS/DatabaseLogic.cs
3. using System.Configuration;

But, if i delete it i get an error, since im using it to get connection string from an xml file.

PR #27, opened 2023-08-20
Inital commit

Ramseis:
I know it's pretty messy. Didn't seem like a meaningful place to split out methods. This is mostly menu navigation without a ton of clear repetition.

### DIFF
--- a/CodingTracker.Ramseis/CodingTracker.Ramseis.sln
+++ b/CodingTracker.Ramseis/CodingTracker.Ramseis.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.7.34003.232
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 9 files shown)


Ramseis:
I'm not sure if I'm not seeing an alternative using call to System.Configuration with a more narrow scope, but it looks like the remaining static analysis error hits basically everyone's project. False positive?

TheCSharpAcademy (changes requested):
@Ramseis Impressive project, well done! I really like that I can change colors. It also checks most requirements. 

I'm going to ask you to check this requirement though, as it's a key learning point of this project: 
image

TheCSharpAcademy:
> I'm not sure if I'm not seeing an alternative using call to System.Configuration with a more narrow scope, but it looks like the remaining static analysis error hits basically everyone's project. False positive?

Yup don't worry about it.

TheCSharpAcademy:
@Ramseis Thanks for making the changes. I'm going to approve the project because the objective here isn't to slow you down, and you demonstrated to have a basic understanding of separation of concerns. However there's still too much going on in Program.CS. You could have put all operations against the database in a separate class and validation on a different one again. And each option in the menu ideally should call it's own method. 

Think of creating small testable classes that do one single thing, which makes the whole thing easier to read, test and debug. This is a good example: https://github.com/TheCSharpAcademy/CodeReviews.Console.CodingTracker/tree/main/CodingTracker.kimfom01/CodingTrackerConsole.

For the next project (Flashcards), given the complexity of the requirements, the code will get messy very quickly if these principles aren't applied. I'll be a bit more strict on that one, but that will be for your benefit 😊😊

PR #28, opened 2023-08-24
initial commit

devhawkz:
I have finished coding tracker with all challenges, but my coding goal doesn't work as it should be. I have found out that query from a database doesn't return any value. Can you look to it and give me some help?


TheCSharpAcademy (changes requested):
@devhawkz Are you able to run the app normally with the latest changes? 

I've checked the repo out and at the moment the connection string is not being correctly populated: 
image

TheCSharpAcademy:
@devhawkz Great project! Very comprehensive reports, well done! 😁

PR #29, opened 2023-08-27
Project finished

maccer989:
Project finished but not familiar with GitHub so hope I have committed correctly


TheCSharpAcademy (changes requested):
@maccer989 Could you please modify the name of thefolder where your project is living to yourGithubName.CodingTracker, as per our conventions? 😊Make sure you test the app and it still runs, you might need to adjust your sln/csproj files.

maccer989 (commented):
Updated to codacity and review comments

TheCSharpAcademy (changes requested):
@maccer989 Thanks for submitting, we're close! It looks like your validation isn't quite working. I'm getting a message saying end time should be after start time, but my original dates are still saved. Also, you need to have the dates for your records as well, so users see their records across time.

image

Also, it would be good to tell the user a record doesn't exist if I try an inexistent id:

image

TheCSharpAcademy:
@maccer989 I'm closing the PR as it's been over 2 weeks, feel free to resubmit it on the platform and reopen the pr when you want to proceed 😁

PR #30, opened 2023-08-29
Project finished

w0lvesvvv:
Haven't done challenges cause i don't have much time and want to skip to other projects

### DIFF
--- a/CodingTracker.w0lvesvvv/CodingTracker.w0lvesvvv/CodingSession.cs
+++ b/CodingTracker.w0lvesvvv/CodingTracker.w0lvesvvv/CodingSession.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CodingTracker.w0lvesvvv
+{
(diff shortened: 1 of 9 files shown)


CodingTracker.w0lvesvvv/CodingTracker.w0lvesvvv/CodingSession.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CodingTracker.w0lvesvvv
+{
+    public class CodingSession
+    {
+        #region DATABASE PROPERTIES
+        public int coding_session_id_i { get; set; }
```
TheCSharpAcademy, line 13: properties should be pascal case

TheCSharpAcademy (changes requested):
@w0lvesvvv well done! we're almost there.

As per our requirements, could you please add the possibility of updating an entry? Also, could you please tidy up coding style according to comments? 😊

TheCSharpAcademy:
@w0lvesvvv Well done! Neat and tidy again. You forgot to display the duration on the table but since I can see it's in the database I'm approving it anyways because I don't want to slow you down. 😊

PR #32, opened 2023-09-12
strawfervor solution

strawfervor:


TheCSharpAcademy:
I'm approving the project, well done! 😁For future reference, get rid of unused usings. You can use this button on the bottom of the code editor window: 

image


image

PR #34, opened 2023-09-15
Updated with fixes to comments

maccer989:


TheCSharpAcademy (changes requested):
@maccer989 I'm still able to log in an incorrect end time. The date is also still not being logged. 😔

image

maccer989:
going to do a different fork asdont understand why my files have not uploaded need to learn some git

PR #38, opened 2023-09-26
Project Finished

Forser:
Hopefully it passes.


Forser:
So it wants me to remove System.Configuration in CodingController / DatabaseManager but if i do that, then ConfigurationManager for Appsettings won't work.

TheCSharpAcademy (changes requested):
@Forser At the moment the program lets me input an end date before a start date and that shouldn't be allowed.
image

Also, I have an interesting problem for you. I get an error trying to view dates probably because the different cultures in our systems. It would be a great exercise to try to solve this issue by making sure reading a date is culture-agnostic. Want to try that?

image

Forser:
Forgot to add a check for Start Date being after End Date, will fix.
I will get fix the culture thing, just got to format it before saving it.

Will push an update today once it's fixed.

PR #42, opened 2023-10-06
Project Finished With Challenges

mekasu0124:
I do apologize ahead of time if I over complicated this project. I made it as efficient as I knew how even with help from various friends. I hope it was what was expected with challenges included.


mekasu0124:
I did not put in the functionality for the challenge of filtering records by weeks, days, year, ascending or descending. I just realized that, but I am happy with the way the project currently sits

TheCSharpAcademy (changes requested):
@mekasu0124 thanks for submitting.

I found a couple of problems. First, the project doesn't build. 

image

Second, it would be good to have a default case in your switch statement. If I input something that is not a switch case the app breaks.

mekasu0124:
> @mekasu0124 thanks for submitting.
> 
> I found a couple of problems. First, the project doesn't build.
> 
> image
> 
> Second, it would be good to have a default case in your switch statement. If I input something that is not a switch case the app breaks.

I do apologize for that. It slipped my mind to check when I separated out my CRUD for goals versus sessions. I have fixed both the namings and the switch case.

TheCSharpAcademy (changes requested):
@mekasu0124 thanks for resubmitting! I'm having these issues: 
1 - When I choose to input manually, there's still a message about stopping the timer. That shouldn't be the case?

2 - You're allowing end time to be before start time. One of the project's requirements is that this shouldn't be allowed

3 - Ideally you should get the date as the user input and not just the time, otherwise I can't input previous dates.

image

TheCSharpAcademy:
@mekasu0124 hey there! Just checking if you still want me to review? I'm asking because you closed the PR 😁

PR #44, opened 2023-10-11
My solution added to the repository.

Paul-W-Saltzman:
Any feedback is welcome.


TheCSharpAcademy (changes requested):
@Paul-W-Saltzman The project is looking great, very comprehensive. 

The only issue is that users should be able to update-delete records. Also, there should be a way to input previous dates, otherwise the functionality is limited. What do you think?

Paul-W-Saltzman:
> @Paul-W-Saltzman The project is looking great, very comprehensive.
> 
> The only issue is that users should be able to update-delete records. Also, there should be a way to input previous dates, otherwise the functionality is limited. What do you think?

Past Dates entries are available under option 1) log coding session.  You may enter today or input a custom date.
image

Update and delete are available under option 3 

image

TheCSharpAcademy (changes requested):
@Paul-W-Saltzman Ok now I saw. Somehow I always expect them in the main menu.

Does updating work for you? If I try this: 
image

I get this: 
image

Paul-W-Saltzman:
Man I tested that early on, but I refactored at several points, but honestly based on the bug I'm not sure how it ever worked.  I had 4 instances of an invalid conversion from a DateTime to a TimeOnly.  I fixed it with the following line of code.

TimeOnly myTimeOnly = TimeOnly.FromDateTime(myDateTime);

Everything should be fixed now.

TheCSharpAcademy:
@Paul-W-Saltzman Well done, the UI looks slick 😁

PR #45, opened 2023-10-11
Finished Base Project + All Challenges!

K-MYR:
I learned so much new about DateTime and SQL during this project! Please be brutal in your review ;) I always want improve my coding skills as much as possible! :) 
As always thank you for reviewing my code! Have a great day!


TheCSharpAcademy:
@K-MYR the project looks great, I love the goals section. I'd usually pass it without comments but since you told me to be brutal 😂

Why do you have to identical paths?
image

Also you could have used a simpler sqlite connection string. That will create the db in the bin folder. Or in your main folder if you change your working directory. 😊

image

K-MYR:
@TheCSharpAcademy i have two strings cuz i set up the App.config in the beginning and the requirements said it should contain the database path AND the connection string. i just forgot to delete the database path in the end cuz i did not it need it😅 i used the full path so i dont have to change the cwd and thought it doesnot matter much.  thanks for the advice, i ll keep it in mind ✌

PR #47, opened 2023-10-13
First try

mariusgrHiof:


TheCSharpAcademy (changes requested):
@mariusgrHiof Thanks for submitting! I couldn't insert a record, what's the expected format?

image

Please avoid unnecessary empty lines (e.g. 1, 11,12,13,14), there are many around the project. [See our code conventions](https://thecsharpacademy.com/article/58) 
image

mariusgrHiof:
> @mariusgrHiof Thanks for submitting! I couldn't insert a record, what's the expected format?
> 
> image
> 
> Please avoid unnecessary empty lines (e.g. 1, 11,12,13,14), there are many around the project. [See our code conventions](https://thecsharpacademy.com/article/58) image

@TheCSharpAcademy Thanks for the review! The format is day/month/year hours:min i.e 20/10/2023 13:15.

I'm going to go through all the files and unnecessary empty lines.

TheCSharpAcademy:
@mariusgrHiof Thanks! I must have been blind  😂. Project approved! 

Just one thing you can think for the future. If there are no records in the database it would be good to handle it, in case I choose to view records. At the moment I get an error. 

image

mariusgrHiof:
@TheCSharpAcademy Thanks for the review! Is this a good way to handle it or is it a better way to handle it in terms of best practice?

image

PR #48, opened 2023-10-16
Add complete project

dnym:
Full documentation at [my personal repository](https://github.com/dnym/tcsa04).


TheCSharpAcademy (changes requested):
@dnym thanks for submitting, it's looking great.
One requirement isn't being met. End date shouldn't be before start date:

image

And this isn't a requirement, but a personal preference. I think using verbatim strings plus string interpolation is way cleaner so you can get rid of all the quotes and slashes. 

image

dnym:
@TheCSharpAcademy Thanks for reviewing! The second image is the same as the first, so I'm not sure what specifically you referred to. The worst I could find were the SQL statements full of escaped quotes, so I converted them into raw strings.

TheCSharpAcademy:
> @TheCSharpAcademy Thanks for reviewing! The second image is the same as the first, so I'm not sure what specifically you referred to. The worst I could find were the SQL statements full of escaped quotes, so I converted them into raw strings.


Yup, that was the one. For future reference, you could use summaries instead of comments when you need explanations like these. This way it can be easily visualized when hovering over a method.
image

PR #49, opened 2023-10-22
Initial

UgniusFalze:


TheCSharpAcademy:
@UgniusFalze  Well done! Project approved!  😊

Just one consideration. It's generally not considered good practice to use exceptions for control of the flow of your program. Exceptions are expensive so they should be left to handle .... exceptions. Input errors aren't exceptions as we know they'll occur so they'd be better handled by other cheaper control flow tools (loops, if/else switches)

image

PR #54, opened 2023-11-20
Finished(?)

MicVerg:
#RoastMe


TheCSharpAcademy (changes requested):
@MicVerg Thanks for submitting! 😁 The project looks good, except that as per requirements, end date can't be before start date:

image

PR #57, opened 2023-12-12
Coding tracker + challenges complete

jkjones98:


jkjones98:
@TheCSharpAcademy I'm not sure what the standard procedure around issues with Codacy is so apologies if this isn't the right way about getting this checked! Codacy is failing the checks because of the following using statements(see screenshots below), the problem is that they are in use so I can't remove them without breaking my app. 

Is this a bug with Codacy or is it because the ConfigurationManager code has a "possible null reference assignment" and I need to ensure that it can never be null before committing and starting a pull request?

Screenshot 2023-12-14 at 06 37 41
Screenshot 2023-12-14 at 06 39 40
Screenshot 2023-12-14 at 06 40 12
Screenshot 2023-12-14 at 06 40 33

TheCSharpAcademy:
@jkjones98 Don't worry about codacy, sometimes it acts funny. I'd just like to say I'm very impressed with your work. You thought of everything: Providing success messages, filtering, allowing user to update just one property and creating a menu for it, all validations, etc. Well done! Keep up the excellent work.

If I could suggest one (minor) thing for better readability, it would be to reduce the indentation by removing brackets. You don't need brackets for namespaces or using statements anymore. Example of your code formatted: 
image

jkjones98:
Thank you for the feedback I really appreciate it! Definitely wouldn't have been able to do all of that without the mentorship of the academy, this has been a massive help for me in my coding journey.As for the indentation i'll try and format it like that for the flashcards project that im on now, think that is an old habit from when i learnt in uni hahaOn 22 Dec 2023, at 06:29, The C# Academy ***@***.***> wrote:﻿
Closed #57.

PR #59, opened 2023-12-19
complete

Speedierone:


TheCSharpAcademy (changes requested):
@Speedierone I think there was a general misunderstanding of the purpose of these tests. You should be checking if the validation works. An example of a test name would be: GivenInvalidDate, return false. Then pass into that test several possible invalid date strings (example: 14/14/2022, 32/01/2023, etc). 

You might need to rework your project so the validation is isolated into its own method. Which will demonstrate how unit tests drive the development. You end up writing code that's more modular and hence more testable.

Speedierone:
@TheCSharpAcademy Im hoping that the corrections I have made are now correct. If not please let me know :)

TheCSharpAcademy (changes requested):
@Speedierone That's it, now the validation tests are isolated! Well done. 🤓

Just to make it perfect, have a look at how you can run the same test with different data and apply it to your project: 
image

TheCSharpAcademy:
@Speedierone Let me know if you have questions 😁

Speedierone:
@TheCSharpAcademy I think i understand. I added test to make sure that the field is not blank, the field has one blank space, the field is one letter and lastly one number. I can see that the example you showed me had the tests in a different format. Would that matter in this case or is it just the case of what people prefer to do?

TheCSharpAcademy:
@Speedierone I'm going to approve because I don't want to slow you down, but in real life you want to use the format I showed. The reason being: when you run tests like that, **each variation is a test case**, which makes it easier to identify issues. The way you're running the tests has **several cases are inside a single test**. If a test fails I have to do further digging to find out what part of the test fails and that's not good practice.

PR #62, opened 2024-01-02
Ibrahim Gilani coding tracker

ibigilani:
I had a reports functionality implemented but it kept messing up when i ran it anywhere besides my machine so i removed it.
I believe the issue had something to do with building my project in the working directory as i forgot to do that when i first ran my application, feel free to be harsh in your review


TheCSharpAcademy:
@ibigilani Well done! Project approved! 🤓 There's just one thing I'd add: When asking for the user to input an Id (such as when deleting or updating), you have to show a list of records, otherwise they'll either need to have memorized the id or go back to view records. 

You could also find a solution for the users not to have to input both dates again when the end date precedes the start date. 

image


But overall great work 😁

PR #63, opened 2024-01-03
Completed. Only last challenge skipped. Not understood clearly.

Doc415:


Doc415:
I ll add the last challenge

PR #65, opened 2024-01-09
finished project

martinka-h:


TheCSharpAcademy (changes requested):
@martinka-h Hey there! Thanks for submitting. It's looking good except that you should validate for end dates being before start dates:

image

martinka-h:
Thanks for the review.

I fixed it it a way. Hope it's okay now:)

PR #66, opened 2024-01-14
Coding Tracker

Sakuraxx:
Finished all basic functions.

### DIFF
--- a/CodingTracker.Cactus/App.config
+++ b/CodingTracker.Cactus/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+	<appSettings>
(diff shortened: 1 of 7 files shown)


TheCSharpAcademy:
@Sakuraxx Well done! I really like that you've used caching in this project. I have to add it as a requirement somewhere. 

One thing I recommend is that you add more separation in your code. Divide it into related chunks otherwise it will quickly turn into spaghetti. Example: 

image

I would do something like this: 
image

Sakuraxx:
> @Sakuraxx Well done! I really like that you've used caching in this project. I have to add it as a requirement somewhere.
> 
> One thing I recommend is that you add more separation in your code. Divide it into related chunks otherwise it will quickly turn into spaghetti. Example:
> 
> image
> 
> I would do something like this: image

Thanks! Done it.

PR #67, opened 2024-01-19
submit for review

frockett:


frockett:
I missed that! Just added and should work correctly now.

frockett (commented):
Should be fixed! I added an if statement in the input validation class to make sure the start time isn't after the end time.

PR #68, opened 2024-02-04
first commit for my code tracker

Mohamed-AbdElhamed99:


stevietv (changes requested):
Nice work! I've taken a look over your project and there are a couple of things that need to be tweaked and fixed up to meet the requirements.

* Validations - don't allow the user to input an end time before the starting time
* Unused code - there are some unncessary usings to be removed, and also a lot of usages of `this.` where its not required. VS shows these in gray letting you know they are unneeded and make for harder to read code.

You also went above and beyond attempting a challenge, great work. I have a couple more comments that don't necessarily mean you didn't mean the challenge but are things that could be improved:

* advise the user whether its a start time or end time they are entering
* on the stopwatch make sure the time doesn't overwrite the current line
image
* Your showing seconds in the output table but always excluding them, could consider just trimming them away also
* You've jumped between `var` and `int` throughout. Both are entirely valid, but try to be consistent

Mohamed-AbdElhamed99:
@stevietv Thank you for reviewing my code 
I added validation function to prevent user from entering end time before start time.
Also i removed unneeded usings

Mohamed-AbdElhamed99:
@stevietv I don't understand your point about 
on the stopwatch make sure the time doesn't overwrite the current line
image

stevietv (approved):

stevietv:
@Mohamed-AbdElhamed99 I've checked the rest and its looking great! I'm gonna approve this one for you.

Watch out next time that your namespaces are all aligned, currently program.cs has an incorrect namespace.

PR #69, opened 2024-02-09
Update missing by design :P

Buutyful:


TheCSharpAcademy (changes requested):
@Buutyful Thanks for submitting! Here are a few bugs/suggestions: 
1 - It would be good to simply have the menu displayed, without the user have to type info.
2 - With Spectre Console all you need is the menu selection
3 - It's good to clear the console after each operation/selection to avoid cluttering
4 - You're allowing users to input an end date earlier than a start date, that violates a requirement of the project
5 - I'm getting an error when displaying a new record, probably because dates in my system are different than yours. It would be good to enforce one date format only so it's independent of culture: 
image

Buutyful:
Hey, thx for the feed back! 
-Changed menu navigation (i had forgot it before)
-Date time parsing more flexible now and no longer possibile to insert start > end

PR #71, opened 2024-02-12
Project finished

Mateusz-Platek:


TheCSharpAcademy (changes requested):
@Mateusz-Platek Thanks for submitting, the project works fine and all requirements are met but there's something that need to be addressed. The way you're creating your SQLite connection is quite problematic. 

You essentially have one instance of the connection for the lifetime of the app. It's generally advised to create an instance per request. This is not the case in your app, but having one connection will create locks in your db in a multithreaded environment, where multiple users try to access the db simultaneously. 

Also, the code does not handle exceptions, which means if an exception occurs after the connection is opened but before it is closed, the connection might remain open indefinitely, wasting resources. It's recommended to use try-catch-finally blocks or the using statement in C# to ensure the connection is closed even if an exception occurs.

Let me know if you have questions.

PR #72, opened 2024-02-12
project finished

Jinboi:


TheCSharpAcademy:
@Jinboi Well done! For future project, please make sure you do some housekeeping before submitting to get rid of unused code, remove unnecessary empty lines and (optional), include a colon after the namespaces so it removes a level of indentation. This class for example would benefit from those 3 changes:

image

Jinboi:
Thank you sir!
Will do housekeeping before submitting going forward.
-Will remove unused codes
-Will remove empty lines
-Will add a semi-colon on the namespace

Also, I wasn't sure too much about validation.
In real projects, I would need something like try-catch validations, correct?
Maybe this is something I can re-visit and improve upon.

Thank you and happy wings Wednesday Hahaha

PR #75, opened 2024-02-22
Complete Coding Tracker project

Dejmenek:


stevietv:
Hi @Dejmenek - thanks for your submission. I've picked this up to review! Nice work on also going for the challenges also!
Right now I don't see a way for a user to manually log a coding session, which is one of the requirements of the project: 

> The user should be able to input the start and end times manually.

Could you take a look at providing this functionality?

A couple of other things I noticed where there could be improvements:
* reset the timer after a coding session is finished, currently starting a new one keeps the old duration
* clear the console when returning to the main menu, it fills up rather quickly with lots of tables

Dejmenek:
Hi @stevietv! I've fixed everything you pointed out and also implemented the mentioned user story👍

stevietv (approved):
thanks for the updates @Dejmenek 
All the requirements are met, well done!

One tip from me would when choosing to delete a session to first show the table.

PR #76, opened 2024-02-23
Coding Tracker project complete

nwdorian:


stevietv:
Hi @nwdorian - great work here, its a really well rounded app where you've also met all the challenges! Awesome!

great way of seeding the database also with Faker, I liked that!

I'm going to approve the submission, though I did come across one issue that you could consider fixing up. When editing a record its possible to set the start date to before the end date, which also causes the duration to be a incorrect

image

stevietv (approved):

nwdorian:
Hi, thanks for the feedback! I added validation for updating the start time of a record. 
image

stevietv:
> Hi, thanks for the feedback! I added validation for updating the start time of a record. image

cheers! Its a little tricky the way its done now. It works for most cased, but what if the user wants to move the coding session forward to a later date, in this way they first have to edit end date, then start date. because they can't set the start date after the end date.
A better way to validate might be to first validate it the `case UpdatingSelection.SaveChanges`?

Its not critical for this project at all, just a bit of polishing!

PR #77, opened 2024-02-25
CodingTracker challenge project

Rxxyxd:
Bit of a learning curve this one was. Let me know if i couldve done anything better:)


CodingTrackerLibrary/Controller.cs
```diff
@@ -0,0 +1,56 @@
...
+                {
+                    conn.Open();
+                    string insertQuery = "INSERT INTO records (DateStart, DateEnd, Duration) VALUES(@DateStart, @DateEnd, @Duration)";
+                    conn.Query(insertQuery, CodingRecord);
+                }
+            }
+            catch (SQLiteException ex) { Console.WriteLine($"SQLiteException Error: {ex.Message}"); }
+            catch (Exception ex) { Console.WriteLine($"Error: {ex.Message}"); }
+        }
+        public static IEnumerable<dynamic> ReadData(string connectionString)
```
Doc415, line 38: According to the requirements you cannot return dynamic object when reading from database.
It should return
IEnumarable<Model> or List<Model> (in your case)

"When reading from the database, you can't use an anonymous object, you have to read your table into a List of Coding Sessions."

Doc415 (changes requested):
Hi there. This is a good start for this project. I have marked and left comments for improvement. After you complete the changes, i ll review it again. 

ps: You can remove all uncessary usings by Code clean up in Vs
image

Rxxyxd:
Hows that?

Doc415 (changes requested):
Thanks for your collabration. Your code looks much better.
Still the automatic session does not save to database. It is failing in the validation.
Take a look at the image below. 
image

Doc415 (approved):
Good to go:) nice work.

PR #79, opened 2024-03-03
Coding Tracker Submission

Fennikko:
This is my submission for the Coding Tracker. I look forward to your feedback!


stevietv:
Hi @Fennikko - firstly, WOW! What a really nicely put together app, its all beautifully present and functional.

There is one thing missing though, the first requirement of the project is to have the same requirements as the Habit Tracker. That means the app should also allow the user to also Edit and Delete their records. Could you go back and include these two functions?

Fennikko:
@stevietv Thank you for the kind words! I have added the requested features.

stevietv (approved):

stevietv:
Nice work, and fully approved! Keep this up!

A small tip from me as a user experience, using the thread.Sleep still passes an input onto the next menu, so you can accidently begin a second study session while the UI displays `1 session added` - you could consider here waiting with a console.readline for a user input to return to the menu.

looking forward to seeing your next project!

PR #81, opened 2024-03-15
Project Submission

obitom67:


Doc415 (changes requested):
hey there. There some unused "usings" in the code and there is an error so the code does not run. Can you fix the issue please.
image

Doc415 (changes requested):
..

CodingTracker.obitom67/DBHandling.cs
```diff
@@ -0,0 +1,141 @@
...
+                tableCMD.CommandText =
+                    @"CREATE TABLE IF NOT EXISTS CodingSessions (
+                        Id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        StartTime TEXT,
+                        EndTime TEXT,
+                        Duration TEXT
+                        )";
+
+                tableCMD.ExecuteNonQuery();
+                connection.Close();
```
Doc415, line 33: He i think you are using Ado.Net here instead of Dapper. 
Can you switch to Dapper alternative for create table query please.

CodingTracker.obitom67/DBHandling.cs
Doc415, line 1: You are also using Ado.Net here. Please use Dapper alternative.

Doc415 (changes requested):
You app is working fine. I marked the methods to be switched to Dapper instead of Ado.net

Suggestions:
User interface does not prompt user for invalid format input. It should be a nice touch to promt user what is going wrong.

CodingTracker.obitom67/DBHandling.cs
obitom67: I saw this documentation for the Reader on Dapper was this so I don't know what else I should be doing for it

Doc415: > 
> 
> I saw this documentation for the Reader on Dapper was this so I don't know what else I should be doing for it

image

this one is much easier with less code

Doc415: > 
> 
> I saw this documentation for the Reader on Dapper was this so I don't know what else I should be doing for it

Try this one, much easier with less code

image

Doc415:
Thanks for collaboration.

Doc415 (approved):

PR #83, opened 2024-03-21
Submission CodingTracker

antonderegt:
This was a very fun exercise. Please give me any feedback you can think of!

### DIFF
--- a/CodingTracker.antonderegt/ActionType.cs
+++ b/CodingTracker.antonderegt/ActionType.cs
@@ -1,0 +1,11 @@
+namespace CodingTracker;
+
+public class ActionType
+{
+    public const string StartSession = "Start session";
+    public const string EndSession = "End session";
+    public const string EditSession = "Edit session";
+    public const string DeleteSession = "Delete session";
+    public const string ShowSessions = "Show sessions";
+    public const string ExitProgram = "Exit program";
+}
--- a/CodingTracker.antonderegt/App.config
+++ b/CodingTracker.antonderegt/App.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+
+<configuration>
+    <appSettings>
+        <add key="connectionString" value="Data Source=codingTracker.sqlite" />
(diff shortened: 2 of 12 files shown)


frockett:
Hi! 

I noticed a couple of things that I think could be changed. First, I noticed that to time the session, the incomplete session gets entered into the database with a minimum DateTime value. This creates some unexpected results if the user tries to look at records before ending the session: 

image

Second, due to the way the session ID is retrieved by start time, the program throws an exception and crashes if you input the same starting time for two different sessions:

image

Once these things are addressed I think it's good to go!

frockett (changes requested):
Hi! 

I noticed a couple of things that I think could be changed. First, I noticed that to time the session, the incomplete session gets entered into the database with a minimum DateTime value. This creates some unexpected results if the user tries to look at records before ending the session: 

image

Second, due to the way the session ID is retrieved by start time, the program throws an exception and crashes if you input the same starting time for two different sessions:

image

Once these things are addressed I think it's good to go!

antonderegt:
Hey @frockett! Thanks for your comments. It helped me make the code a bit simpler, by getting rid of two methods. Summary of my changes: 
I only save a session to the database when it's done. This way I don't have to check in the database to find the running session and the method with the error could be removed. Also when editing a session I used a new variable instead of using the variable of the running session, to be able to edit sessions while a session is running.

Hope this fixes your comments. Thanks again!

frockett (approved):
Hi @antonderegt , everything looks good to me and I'll approve the project. I think the way you adjusted the SessionController is great and a big improvement.

For similar projects in the future, I might suggest preventing the user from creating a session with the same or overlapping times, but that's just a suggestion, it's not necessary in this case. 😉

You're also helping me by allowing me the opportunity to practice my reviewing and testing skills, so thank you for that!

antonderegt:
Thanks @frockett! I'll keep your suggestion in mind. Error (or user) handling can get complex fast it seems.

I'll leave some more bugs for you next time 😝

PR #85, opened 2024-03-24
CodingTracker.JaegerByte

JaegerByte:
My version of the CodingTracker. This one was really fun to make.
Any tips or feedback would be greatly appreciated!

### DIFF
--- a/CodingTracker.JaegerByte/Program.cs
+++ b/CodingTracker.JaegerByte/Program.cs
@@ -1,0 +1,193 @@
+﻿using Spectre.Console;
+using Microsoft.Data.Sqlite;
+using Dapper;
+using System.Configuration;
+using System.Collections.Specialized;
+
+namespace CodingTracker.JaegerByte
+{
+    enum MenuOption {insertSession, deleteSession, updateSession, viewAllSessions, CloseApplication}
+    internal class Program
+    {
+        static Validation validationHandler = new Validation();
+        static string connectionString = ConfigurationManager.AppSettings.Get("ConnectionString");
+
+        static void Main()
+        {
+            CreateTable();
+
+            while (true)
+            {
+                Console.Clear();
+                MenuOption selectedOption = GetOption();
+                switch (selectedOption)
+                {
+                    case MenuOption.insertSession:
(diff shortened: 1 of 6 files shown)


CodingTracker.JaegerByte/Program.cs
```diff
@@ -0,0 +1,191 @@
...
+            return AnsiConsole.Prompt(selectionPrompt);
+        }
+
+        static void InsertSession()
+        {
+            string startTime = AnsiConsole.Ask<string>("Insert Start Time (Format HH:mm)?");
+            string endTime = AnsiConsole.Ask<string>("Insert End Time (Format HH:mm)?");
+            if (validationHandler.CheckDateInput(startTime, endTime))
+            {
+                using (SqliteConnection connection = new SqliteConnection(connectionString))
```
frockett, line 82: You could refactor the methods that do CRUD operations to a separate file/class

frockett (changes requested):
Hi again @JaegerByte ! I promise I won't be the only person to review your projects, in case you're tired of seeing me twice haha. 

There are two requirements that I think need to be adjusted for. Otherwise, I think this one does a good job with things like validation and is also a clear improvement form the habit tracker in many ways.

The first requirement is this one: 
image
Refactoring the logic that accesses your database and the logic that gets input from the user into their own classes could help improve the design of the program. In general, you can imagine that each class should have a single responsibility. For example, Validation checks if inputs are valid integers or strings, UserInput gets input from the user and sends it to Validation for validation, then sends it on to Controller/Repository, which turns that input into the data model and feeds it into the database. Apologies if I'm explaining something you're already familiar with! But that's what this requirement is talking about. 

The second requirement is a smaller adjustment: 
image
The user should be able to input a date as well as a time.

Also, I want to thank you for the chance to practice my reviewing and testing skills! Let me know if anything is unclear.

frockett (approved):
Looks good to me! I'll approve it

JaegerByte:
> Looks good to me! I'll approve it

Thank you for your efforts! I think my code would be much more structured if I reprogrammed the project from scratch now, with a bit more experience :D

PR #86, opened 2024-03-28
Completed project

caloc-nz:


VocalNight (approved):
Very nice. I liked the decision to only add timer based records if they go above 1 minute.
The only thing i would like is to see the timer when it is counting up, right now you only see it when you pause/stop the timer

image

But overall very good job, i will be approving it.
Good luck on your next projects.

PR #88, opened 2024-04-07
Coding Tracker Project Complete

GetTeched:
Thank you in advance for taking your time in reviewing my code!


GetTeched:
Codacy error with the following
image

If I remove that I get an error here.
image

GetTeched.Console.CodingTracker/InputValidation.cs
```diff
@@ -0,0 +1,98 @@
...
+            return totalDuration = String.Format("{0:00}h {1:00}m {2:00}s", hours, minutes, seconds);
+        }
+        else if (minutes > 0)
+        {
+            return totalDuration = String.Format("00h {0:00}m {1:00}s", minutes, seconds);
+        }
+        else
+        {
+            return totalDuration = String.Format("00h 00m {0:00}s", seconds);
+        }
```
antonderegt, line 88: For readability you can get rid of the else. 

I see you already did that in the method `DateTimeParse`.

GetTeched.Console.CodingTracker/RandomSeed.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    internal void GenerateRandomData()
+    {
+        Random random = new Random();
+
+        int loggedDays; int day; int month; int year = 23; int hours; int minutes; int seconds; string startDate; string endDate; string duration;
+
+        loggedDays = random.Next(30, 260);
+
+
```
antonderegt, line 23: This class has a few redundant empty lines. For example on this line.

GetTeched.Console.CodingTracker/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿using coding_tracker;
+using System.Configuration;
+
+bool testMode = true;
+
+string connectionString = ConfigurationManager.AppSettings.Get("ConnectionString")!;
+DatabaseManager databaseManager = new();
+SessionController sessionController = new();
```
antonderegt, line 8: In this file you get the `connectionString` and in the `SessionController` you also get the `connectionString`. File IO relatively takes a lot of time, so, for performance it would be nice to only load the `connectionString` once.

antonderegt (commented):
Hey @GetTeched really good job! I see you put a lot of effort in it by implementing not only the requirements but also the extra challenges. I did have some issues when initially running it though. You used both System.Data.SQLite and Microsoft.Data.Sqlite. And the random seed method sometimes gives error messages like this:
```
Unhandled exception. System.FormatException: String '20-03-23 05:01:35' was not recognized as a valid DateTime.
   at System.DateTimeParse.Parse(ReadOnlySpan`1 s, DateTimeFormatInfo dtfi, DateTimeStyles styles)
   at System.DateTime.Parse(String s)
   at coding_tracker.RandomSeed.GenerateRandomData() in /Users/anton/Documents/tcsa/Reviews/CodeReviews.Console.CodingTracker/GetTeched.Console.CodingTracker/RandomSeed.cs:line 33
   at coding_tracker.UserInput.Testing(Boolean testMode) in /Users/anton/Documents/tcsa/Reviews/CodeReviews.Console.CodingTracker/GetTeched.Console.CodingTracker/UserInput.cs:line 22
   at Program.<Main>$(String[] args) in /Users/anton/Documents/tcsa/Reviews/CodeReviews.Console.CodingTracker/GetTeched.Console.CodingTracker/Program.cs:line 13
```

When you application starts it shows a few warnings. I always try to treat warnings like errors. That way you are certain your application is robust and will be able to handle everything the users will try to throw at it.

Your methods are nice and short. Your naming is descriptive. Very well done 🥳 🎉 ! I left a few comments in the code, but those are just nitpicky things. You don't have to fix them, I'm going to approve this project! I hope you can keep up the determination to do all the extra challenges, that was really nice to see 😄

GetTeched.Console.CodingTracker/InputValidation.cs
GetTeched: That was silly of me, no idea how I missed that

GetTeched.Console.CodingTracker/Program.cs
GetTeched: I guess I could have combined  `DatabaseManger` and `SessionController` actually don't think it was necessary to split those, then I would only need to call `connectionString` once.

PR #91, opened 2024-04-16
Coding Tracker Completed

Aditya-083:


Aditya-083:
@antonderegt  please have a look.

antonderegt:
> @antonderegt please have a look.

Hi @Aditya-083, I'm sorry, I don't have time for a good review at the moment. Your submission is visible in the review portal so someone will get to it. Hang tight.

stevietv:
Hi @Aditya-083 - thanks for your submission.  Overall great work and meeting all the requirements. Good Job :)

Here are a few suggestions for you:

* variables declared within a method should follow camelCase not PascalCase - you can check [this](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions) out for reference 
* when choosing `Delete Record` it would be wise to present the user the list of records available to delete.
* You may also want to consider using a [ConfigurationManager](https://learn.microsoft.com/en-us/dotnet/api/system.configuration.configurationmanager?view=dotnet-plat-ext-8.0) instead of the appsettings.xml - you could reduce your code to get the connection string to 1line instead of 5
* you might consider presenting your records also in a Spectre Table instead of text, it can be easier to read the output

One more important point of feedback:

You submitted this project not in a subfolder, next time could you put your work into a folder following the naming structure "CodingTracker.Aditya-083" so that its easily traceable to your user. That makes it easier when reviewing multiple submissions.

Aditya-083:
@stevietv thanks for the review I will keep these points in my mind when raising a review next time.

PR #93, opened 2024-04-23
Fixed reports and added order by desc/asc

Mo3ses:
Hi I came back, probably I will not do the last two Challenges because I stopped this project a while ago and I confess that I am a little lost in his structuring so I only got the last two Challenges, if everything is correct could leave this as concluded on the site?

### DIFF
--- a/CodingTracker.Mo3ses/App.config
+++ b/CodingTracker.Mo3ses/App.config
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+
+<configuration>
+	<appSettings>
+		<add key="dbconnectionString" value="Data Source=Coding-Tracker.db" />
+		<add key="en" value = "Hello Word"/>
+		<add key="br" value = "Olá Mundo"/>
+	</appSettings>
+</configuration>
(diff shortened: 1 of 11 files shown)


Mo3ses:
so how i should handle theses codacy errors 15 of them is about line length but is on readme.md, one of them is this one.
image
says it is unnecessary but as you can see I am using it.
usingConfiguration
usingConfiguration2

stevietv:
Hi @Mo3ses - thanks for coming back and finishing this one off. I see that you started it quite a while ago when `ConsoleTableExt` was still the requirement. Nowadays we use `Spectre.Console` but since you started this one before the requirement changed I'll let it slide! Do check it out though as Spectre Console is great!

We also recently added the requirement to use Dapper ORM instead of ADO.net - you might always want to look into that at some point.

Theres some things that could be improved, for example you sometimes write `Console.WriteLine` and sometimes `System.Console.WriteLine`-  the System part is not needed as its imported as a default using.

Also, you do a lot of UI presentation within the DbConnnect class in the GetAll Method - you should keep the DbConnect class as a simple class that gets the data from the DB and returns it to the UI for presentation.

All in all, a valid attempt and a great job at attempting two of the challenges. 

Approved ✅

PR #94, opened 2024-04-26
Project completed

gabriel2443:


kronyer (changes requested):
Hey! It's me again:

It looks that different from the previous project you separated the functions, very nice!
So:

1- The first things is that you run the menu before you create the database, you should invert the order, so the database is created
image

The other things are perfect, It's also very neat.

Also, there are minor things that make you code more clean, like putting a ";" after the namespaces, so the {} are not necessary anymore
image
And you should remove the unnecessary blank spaces too (i just see these two in the Model) the other classes are perfect.

gabriel2443:
Hey, thank you for your feedback very much appreciated, I hope now is fine.

kronyer (approved):
It looks perfect now. Well done!

PR #97, opened 2024-05-28
Coding Tracker Console App

N-Endy:
I tried implementing all features to the best of my ability.


frockett (changes requested):
Hi, I'm glad I get to be the one to review your final beginner project. 

I noticed something that isn't a strict requirement, but that I think would make a big difference. When updating a session, the table of sessions isn't displayed. This makes it more difficult for the user to know what they're editing. You might want to print the table and the selected session for a better user experience.

image


Second, and more critical to the completion of this project, is that it is currently possible to update a session with an end time that is earlier than the start time. This should be impossible. Here's an example where I did it: 
image

I notice you do have some validation, but it doesn't seem to be working in every case. For example, here you can see it allowed me to input sessions with negative durations in two different ways. The first one was by update, and the second one was by a normal insertion.
image


Make sure that all dates are validated correctly and the project should be good to go. If you need help looking at how to do that then leave a comment or shoot me a message on discord (@anger_goose in the server) and I'll be happy to give more specific advice, I just figured you might want to give it a shot on your own first 🥳

N-Endy:
Thank you for the feedback. I will make the necessary changes.

N-Endy:
Hello,

Thank you once again for the feedback on my project. I have added
validation to check that endTime isn't earlier than startTime when updating
a session.

For the third feedback on accepting an endTime, if the date is earlier, you
are using a later time for the endTime from the image you shared with me.
You used a start time of "02-01-20 12:00:00" and an endTime of "31-01-20
16:00:00". The endTime is later than startTime and thus would be accepted.

frockett (approved):
Looks good! Just checked everything and the validation catches those cases. I'm going to approve. Congrats on finishing the first section of console projects! 🎉

PR #101, opened 2024-06-08
Coding Tracker app with all challenges.

arashi256:


arashi256:
Codacy check is wrong - the using is required to reference ConfigurationManager.

Doc415 (changes requested):
Hey your project looks great.
Unfortunately there are some parts that needs attention :<

When i try to update or delete first record in the list, it throws exception
image

After i recorded two live sessions, i select "list last 7 days coding sessions" and it shows none of them
image

--suggestion--
it is better to format decimals to 2-3 places
image

--suggestion--
it is not a good idea to let user fiddle with DB
image

arashi256:
Could not reproduce update/delete coding session Exception. Other things fixed + patch to cope with particularly long TimeSpans for coding session durations.

Doc415 (approved):
Nice work. Approving you project.

Doc415:
Nice work . Approving your project.

PR #105, opened 2024-06-30
Finished the project

kjanos89:


kalsson:
Hello @kjanos89,

Sorry for a little late code review. From what I can see right from the start, you are missing a ReadMe file.

- Your project needs to contain a Read Me file where you'll explain how your app works. This comes from the (Habit Tracker) project but it also applies in this project.

kalsson:
image

I'm having a hard time adding any record. You probably need to check that before I can continue.

kjanos89:
Hello, @kalsson ! Thank you for reviewing my code.

After checking the requirements I was also missing the one about reading data from the db into a list. Added that and the readme file as well.
For the records you'll need to add an ending dot, at first I added an example into the "error" string, now I've also added a hint.
Sorry for the confusion, if it isn't good enough I can rework the validation for the input strings and loosen the requirements.
Currently codacy gives me 2 errors, one for the first line and one for the 29th which is empty. As far as I know these are well known issues with it with no workaround so far so I'll let them be, these are also a reason why I'd rather add examples to my feedback strings than add more info for the readmes.
Once again, thank you for your time and effort in reviewing my code.

kalsson:
> Hello, @kalsson ! Thank you for reviewing my code.
> 
> After checking the requirements I was also missing the one about reading data from the db into a list. Added that and the readme file as well. For the records you'll need to add an ending dot, at first I added an example into the "error" string, now I've also added a hint. Sorry for the confusion, if it isn't good enough I can rework the validation for the input strings and loosen the requirements. Currently codacy gives me 2 errors, one for the first line and one for the 29th which is empty. As far as I know these are well known issues with it with no workaround so far so I'll let them be, these are also a reason why I'd rather add examples to my feedback strings than add more info for the readmes. Once again, thank you for your time and effort in reviewing my code.

Hey @kjanos89 !

Thanks for your explanation. There's no problem with you ending with a (.), I'm just used to not doing that. I will continue to evaluate your code now.

As you say, there are a lot of problems with codacy so you can ignore that.

kalsson:
Good work! This is approved. 😁

- Great work with the user feedback for the delete action and the double check to make sure if the user want's to delete a record. 👍

- Greate work with the user feedback for the update record, when the user enters an ID that doesn't exist. 👍

Here is a list of what you can think of for your next project. (Nothing you need to fix in this project):

1. When a user adds a record, it could be nice with a success message if saving to the database went well. This is not a problem as the user can choose to display records.

2. It can be a good idea if the user can, for example, type "exit" to cancel an operation. Currently if I choose to add something I have to complete it or restart the application.

3. When a user want to delete and update an entry, it would have been good to show the user which IDs exist in the database.

4. If I try to delete an ID that doesn't exist, the application jumps back to the menu. You have handle so that the application doesn't crash which is great but try to think about user feedback. Do a check first if the ID exists in the database before trying to delete it.

PR #108, opened 2024-07-06
CodeTracker project for Code Review

AlvinGraham:
Completed all base and challenge requirements.


CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Utilities.cs
```diff
@@ -0,0 +1,116 @@
...
+			throw new ArgumentException("Returning to Main Menu");
+
+		var startDate = Validation.ValidateStartDate(startDateInput);
+
+		var endDateInput = AnsiConsole.Ask<string>("Input End Date with the format: dd-mm-yy hh:mm (24 hour clock), or enter 0 to return to main menu.");
+
+		if (endDateInput == "0")
+			throw new ArgumentException("Returning to Main Menu");
+
+		var endDate = Validation.ValidateEndDate(startDate, endDateInput);
```
javedkhan2k2, line 50: What if the user entered the start date incorrectly. You are only asking for entering endDate.
You can ask the user for both start and end date until both are correct or user cancel the operation.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Utilities.cs
```diff
@@ -0,0 +1,116 @@
...
+
+		return output;
+	}
+
+	internal static DateTime[] GetDateInputs()
+	{
+		var startDateInput = AnsiConsole.Ask<string>("Input Start Date with the format: dd-mm-yy hh:mm (24 hour clock), or enter 0 to return to main menu.");
+
+		if (startDateInput == "0")
+			throw new ArgumentException("Returning to Main Menu");
```
javedkhan2k2, line 41: Throwing an Exception in these cases are not good practice.
Please update this in all places.
Try to display to user " You canceled the operation and Returning to Main Menu" return e.g. Null and update the code accordingly.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/UserInterface.cs
```diff
@@ -0,0 +1,183 @@
...
+
+		Console.WriteLine(responseMessage);
+		Console.ReadKey();
+	}
+
+	private static void UpdateRecord()
+	{
+		var dataAccess = new DataAccess();
+		var records = dataAccess.GetAllRecords();
+		ViewRecords(records);
```
javedkhan2k2, line 140: It will be nice not to show an empty table, especially when updating/Deleting a record.
You can return when there are no records in database.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/UserInterface.cs
```diff
@@ -0,0 +1,183 @@
...
+		table.AddColumn("Duration");
+
+		foreach (var record in records)
+		{
+			table.AddRow(record.Id.ToString(), record.DateStart.ToString(), record.DateEnd.ToString(), $"{(int)record.Duration.TotalHours} hours {record.Duration.TotalMinutes % 60} minutes");
+		}
+
+		AnsiConsole.Write(table);
+
+		Console.WriteLine("Press any key to return to main menu.");
```
javedkhan2k2, line 168: When viewing the Updating/Deleting records, kindly removed the “Press any key to return to main menu.”
When user enter 0, we are back main menu. It will be better to update the messages.
Please type the id of the habit you want to update.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Validation.cs
```diff
@@ -0,0 +1,62 @@
...
+
+	internal static DateTime ValidateEndDate(DateTime startDate, string endDateInput)
+	{
+		DateTime endDate;
+		while (!DateTime.TryParseExact(endDateInput, "dd-MM-yy HH:mm", CultureInfo.InvariantCulture, DateTimeStyles.None, out endDate))
+		{
+			endDateInput = AnsiConsole.Ask<string>("\n\nInvalid date. Format: dd-mm-yy hh:mm (24 hour clock). Please try again\n\n");
+		}
+
+		while (startDate > endDate)
```
javedkhan2k2, line 49: I entered startDate = endDate and it is added to database.
I am also not 100% sure if one should entered 0 duration coding tracking session.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/DataAccess.cs
```diff
@@ -0,0 +1,167 @@
...
+			foreach (var record in records)
+			{
+				record.Duration = record.DateEnd - record.DateStart;
+			}
+
+			return records;
+		}
+	}
+
+	internal List<CodingRecord> GetRecordList()
```
javedkhan2k2, line 78: No need to create a function for List of records
`var records = connection.Query<CodingRecord>(selectQuery);`
You can get list as
`records.ToList()`

javedkhan2k2 (changes requested):
A very well written Project.
To further improve please fix the bugs/issues and commit.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Utilities.cs
AlvinGraham: I'm not exactly sure what you mean here.  I would point out that this method does indeed ask for and validate both startDate and endDate (sequentially) before returning both values as part of a DateTime array.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Utilities.cs
AlvinGraham: Good call, thanks. I was trying to get cute with returning all the way to the MainMenu despite where it is in the call stack.  I see how this is bad practice and have refactored the code so that the method returns to the correct calling code and notifies the user as such.  I admit this was intellectual laziness on my part and was an easy fix.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/UserInterface.cs
AlvinGraham: Acknowledged. I made the changes for Updating/Deleting records and updating/deleting goals.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/Validation.cs
AlvinGraham: Good point.  Fixed and thanks.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/DataAccess.cs
AlvinGraham: I think I see what you are saying here.  In essence I had two methods that did the exact same thing but one returned a List and the other returned an IEnumerable.  I kept the IEnumerable one and just used .ToList() method where necessary.

CodingTracker.AlvinGraham/CodingTracker.AlvinGraham/UserInterface.cs
AlvinGraham: Ha, I inadvertently fixed this while resolving issue #2 above!

javedkhan2k2:
@AlvinGraham approved and good luck with your next Project.

PR #112, opened 2024-07-20
done

kwm0304:


kwm0304:
Codacy is telling me I have an unused using statement but I'm using it right below that

TheCSharpAcademy:
@kwm0304 Well done! Very comprehensive project. It's great that you completed the challenges 😁

Two things: I think there are too many layers. The code you have in the Db Action should be in the repository. At the moment the repository is merely forwarding messages to Db Action, so it's looking a bit redundant.

Also, the user needs to know what's the expected format at all times 🤓
image

Other than that, great project!

PR #119, opened 2024-08-11
Project Completed

PokingOutsideOFF:
This project took more time than i expected. Please provide any suggestions required to make this project more better. The project structure I think is confusing as I wasn't able to decide which class should reference which project. However I hope functionalities of the problem are met.

### DIFF
--- a/CodingGoal/CodingGoal.csproj
+++ b/CodingGoal/CodingGoal.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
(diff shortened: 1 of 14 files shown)


TheCSharpAcademy:
@PokingOutsideOFF  Well done! Project approved! 👌🏻🤓✅ And well done for completing the challenges! 

Ok since this is your 4th project I can start being a bit more strict 😄:

1 - Why so many projects? You could simply create folders. My advice is to always create a monolith and start separating into projects/micro-services as it grows. 

2 - Why not use auto-properties where you can? 40 lines would become 10 😄 :
image

3 - Whenever you have a fixed index I recommend you use an enum. This will improve readability. I modified a few lines of your code as an example: 
image

And congratulations! You got your Olive Green belt!! 🫒🫒

See you in the next project 🤓

PokingOutsideOFF:
Thank you for this feedback 😎
I would keep in mind for the future projects. 
However i still have confusion regarding how to create the project structures.

PR #127, opened 2024-08-24
Finsihed Project

MansoorAZafar:


CodingTracker.MansoorAZafar/CodingTrackerLibrary/Controllers/CodingHabitManager.cs
```diff
@@ -0,0 +1,217 @@
...
+    {
+        Console.Clear();
+        int id = -1;
+
+        switch(chosenSelection)
+        {
+            case MenuSelections.update:
+                DataViewer.DisplayHeader("Updating a Coding Habit");
+                
+                this.UpdateCodingHabit(ref id);
```
chrisjamiecarter, line 25: Remember that you can do 
`id = UpdateCodingHabit();`

and then change the signature of the method to:
`private int UpdateCodingHabit()`

Which is easier to read than:
`this.UpdateCodingHabit(ref id);`

chrisjamiecarter (changes requested):
Hi @MansoorAZafar ,

I am reviewing your app, but when I launch it I am getting the following error:

image

Could you try replicate from your end on a clean/fresh database? As it seems to be the data seeder that is causing the issue.

Let me know if you need any support.

Commit and Push your changes when complete and I will continue my review.

Thanks
@chrisjamiecarter

CodingTracker.MansoorAZafar/CodingTrackerLibrary/Controllers/DatabaseManager.cs
```diff
@@ -0,0 +1,249 @@
...
+    public List<CodingSession> GetAllData()
+    {
+        List<CodingSession> codingSessions = new();
+        using var connection = new SqliteConnection(this.connectionString);
+        connection.Open();
+
+        using var reader = connection.ExecuteReader("SELECT * FROM CodingTracker");
+        while(reader.Read())
+            codingSessions.Add(new CodingSession(reader.GetInt32(0), reader.GetString(1), reader.GetString(2), reader.GetFloat(3), reader.GetString(4)));
+
```
chrisjamiecarter, line 64: You could have done this to simplify your code a lot:
`return connection.Query<CodingSession>("SELECT * FROM CodingTracker").ToList();`

Dapper handles the tedious work of mapping database rows to objects, so you don't need to manually extract data from a SqlDataReader.

chrisjamiecarter (approved):
Hey @MansoorAZafar ,

Great coding tracker application! 🎉
- Good creative use of the Spectre.Console library.
- Good structuring of classes and responsibilities into different files.

**User Experience**
- You could allow the user a way out of adding/updating/deleting an item. By pressing 0 for example.

I will go ahead and mark as approved, keep up the excellent work on the next projects! 😊

Best regards,
@chrisjamiecarter

PR #129, opened 2024-08-30
updates

gafoorraees:
submitting the requested fixes for consideration. thank you.


Coding-Tracker/coding-tracker/.gitattributes
chrisjamiecarter, line 1: File not required. Should be deleted from the Pull Request submission.

Coding-Tracker/coding-tracker/.gitignore
chrisjamiecarter, line 1: File not required. Should be deleted from the Pull Request submission.

Coding-Tracker/coding-tracker/coding-tracker/CodingSession.cs
```diff
@@ -0,0 +1,45 @@
...
+    public DateTime Date { get; set; }
+    public DateTime StartTime { get; set; }
+    public DateTime EndTime { get; set; }
+    public TimeSpan CalculateDuration
+    {
+        get
+        {
+            if (EndTime < StartTime)
+            {
+                return (EndTime.AddDays(1) - StartTime);
```
chrisjamiecarter: This code is redundant as your validation logic doesn't allow `EndTime` to be less than `StartTime`.

chrisjamiecarter (changes requested):
Hey @gafoorraees ,

First off, excellent work on another application! 🎉 It’s great to see your progress.
- Great use of the Dapper ORM to aid your data queries.
- Nice use of ISO 8601 for Date input format.
- Perfect code styling.

As always, I have a performed a fair review and have few suggestions that could help your growth as a developer:

**Solution Structure**
Re-read the [Code Reviews](https://thecsharpacademy.com/article/52/code-reviews) academy page. Specifically step 3.
This will make it easier for code-reviewers to pull down and review your code in the future.
When submitting your code review pull request, all your solution files should be contained inside a new folder as per below (currently it is coding tracker inception!):

```
// Current.
.
└── Coding-Tracker/
    └── coding-tracker/
        ├── coding-tracker/
        └── coding-tracker.sln

// Correct.
.
└── CodingTracker.gafoorraees/
    ├── CodingTracker/
    └── CodingTracker.sln
```

And also, any .git* files are not needed, as they already exist in the parent project directory.


**Style Guide**
- Use the simpler "Using Declaration" syntax to save an indentation level against the "Using Statement" syntax:
```
// Using Statement.
using (var connection = new SqliteConnection(connectionString))
{
    // Code here.
}

// vs

// Using Declaration.
using var connection = new SqliteConnection(connectionString);
// Code here.
```

**Other Remarks**
- Try to think about separating the concerns of each file. For example, you currently open a connection to the database and then asking questions to the user and waiting input. In real life this would cause massive issues. Why not ask the questions first? Get the answers. Validate. Then open a connection and add to the database. 
- You have created the `UserInput.cs`/`Validation.cs` files. Make sure you are using them 😊. I have seen repeated code in methods that are not making use of the code you have written in these files. This will help with the DRY principle too. Do not Repeat Yourself, i.e. don't write the same code twice.
- You can enter a coding session in the future.
- When updating a record, the `EndTime` can be greater than the `StartTime` but not when Inserting (enter a start value > end and then press enter to keep default).
- What about if I am a night owl and I am coding past midnight? Currently on manual Insert it is not possible. (but possible on update as per above point. And presumably possible on a live coding session). Why not ask for a date on the `EndTime` too?
- When Inserting a record, there is no option to cancel.
- When Deleting a record, the old table is still present. Maybe clear this to show the update/deletion has happened successfully.

**Bugs**
The above are just suggestions. But unfortunately, I did spot a bug within your application that will need to be fixed:

1. NULL Description:
I noticed a small issue with the Insert Coding Session functionality.
It looks like when you press enter at the description. It prompts for a valid description but the next results are not retained and the INSERT query throws an exception.

image

Please revisit this issue and once fixed, Commit and Push your code, and I will complete my review.

Feel free to reach out if you have any questions or if you'd like to collaborate further on any of these points.

Thanks,
@chrisjamiecarter

gafoorraees:
Hi Chris,

Thank you so much for taking the time to look through the code and for your insightful comments. I greatly appreciate it.

I fixed the description bug that you caught such that the table should retain the results after the initial re-prompting for a valid description.

For separation of concerns, I changed it such that I prompted the user for input and then opened the connection. However, I had to change it back for the update and delete methods because I noticed that I was still displaying the session data to the user even when there was no data. I will think about how to fix this. Thank you for bringing it to my attention.

For inserting records, the user can now return to the main menu during each insert prompt.

For deleting a record, the table is now cleared after a successful deletion.

I did not fix the coding past midnight issue, but I understand how adding an EndTime method and column would solve that problem. I will keep it in mind as I go forward. I also will keep in mind the DRY principle as I go forward. 

Thank you!!

chrisjamiecarter (approved):
Well done @gafoorraees ,

Perfect updates! Fantastic job on this app. I hope that my review helped in guiding your growth as a developer.
I will go ahead and mark as approved, keep up the excellent work on the next projects! 😊

Best regards,
@chrisjamiecarter

PR #130, opened 2024-09-03
FINISHED

Mattias1610:


Paul-W-Saltzman:
Hello,  could  you please change your directory name from Mattias1610\Coding-tracker to Mattias1610.Coding-tracker the backslash is not a valid character for a file or directory at least in windows.  Once you make that change, I'll get it reviewed.  Thanks.

Paul-W-Saltzman:
Thank you for making the change to the file name.  I took an in depth look at your program and it looks like you are really getting a good handle on your C# code.  I really like your enter date then time model.  I think this makes it easier for the user to enter the information.  This stuff is not easy and you are doing well so far.  On this assignment you missed a couple of the requirements though.  

- You'll need to create a configuration file that you'll contain your database path and connection strings. 
- You'll need to create a "CodingSession" class in a separate file. It will contain the properties of your coding session: Id, StartTime, EndTime, Duration.  (Currently you are displaying directly from the database you need to create a class and load the data as objects.  You will need to change your database to store the start time and end time.)
- When reading from the database, you can't use an anonymous object, you have to read your table into a List of Coding Sessions. (see above)

The main piece to remember is that C# is an Object oriented language.  You should be loading the information from the database into objects and then interacting with the database.   Also Consider doing a little more validation.  I was able to enter an end time that was chronologically before the start time.  

Let me know if you need any guidance.  Keep Coding!

PR #134, opened 2024-09-14
Finished

JvR-Hannes:


README.md
chrisjamiecarter, line 1: Hey @JvR-Hannes 👋 ,

Could you move this file into your project folder please?

NOTE: You already have a README.md in you project folder, but it is empty. So maybe copy/paste contents and then delete this file?

Either way, once done: commit and push the change (which will update this PR).

I will then begin my review.

Thanks
@chrisjamiecarter

chrisjamiecarter (changes requested):
Please make changes as per comment.

chrisjamiecarter (changes requested):
Hey @JvR-Hannes,

Unfortunately I have found a couple of bugs in your code...

When viewing all records on a new database...

image

I can workaround by adding a record. But this needs fixing as it fails the requirement carried over from [habit-logger](https://thecsharpacademy.com/project/12/habit-logger):
- [ ]  You should handle all possible errors so that the application never crashes.


You are currently failing the [requirement](https://thecsharpacademy.com/project/13/coding-tracker):

- [ ]  You should tell the user the specific format you want the date and time to be logged and not allow any other format.

If you are requesting `yyyy-MM-dd HH:mm:ss` you should only accept this. Read up on TryParseExact

image


See the [tips section](https://thecsharpacademy.com/project/13/coding-tracker):

- Don't forget the user input's validation: Check for incorrect dates. What happens if a menu option is chosen that's not available? What happens if the users input a string instead of a number? **Remember that the end date can't be before the start date**.

Currently you allow the end date to be before the start date, see ID: 9.

image

Please could you fix, let me know if you need any support.

Don't be disheartened, the project is nearly complete! Just a few issue to iron out!

Thanks
@chrisjamiecarter

JvR-Hannes:
Ok, I'll check it out, looks like I didn't add a closing tag there, will
fix and update repo

chrisjamiecarter:
Also. I am not sure and am checking. But I think you also need to provide an update method.

So you can create, read, update and delete a coding session.

Apologies I didn't add to the previous comment.

JvR-Hannes:
Ok yeah, I saw that I completely missed it, I'm not sure what happened that
make me miss it. I have added it and also the closing tag on the no records
found.

Thanks for letting me know

chrisjamiecarter (approved):
Well done @JvR-Hannes on another complete project 🎉

PR #135, opened 2024-09-16
Submission of coding tracker

A-Altemara:


CodingTracker.A-Altemara/CodingTracker/Menus/Menu.cs
```diff
@@ -0,0 +1,307 @@
...
+                    GoalsMenu.AddNewEntry(_goalsDb);
+                    break;
+                case "Edit Existing Goal":
+                    EditEntry(_goalsDb);
+                    break;
+                case "Delete a Goal":
+                    DeleteEntry(_goalsDb);
+                    break;
+                case "View a goal":
+                    var goal = GoalsMenu.GetGoal(_goalsDb);
```
Paul-W-Saltzman, line 120: You have a bug here where the user can exit the GetGoal method and then when the program gets to the ShowProgressToGoal method the goal and sessions are null.

CodingTracker.A-Altemara/CodingTracker/ICodingTrackerDb.cs
Paul-W-Saltzman, line 1: You are very heavy on comments in this file.  The goal is to name your methods, objects and write your code so clearly that you don't need comments.  On the other hand comments are free.  There is a balance to find here.

CodingTracker.A-Altemara/CodingTracker/CodingDb.cs
```diff
@@ -0,0 +1,262 @@
...
+                StartTime = randomStart,
+                EndTime = randomEnd,
+                Duration = duration
+            };
+            counter--;
+            _dbConnection.Execute(insertQuery, entry);
+        }
+    }
+
+    /// <summary>
```
Paul-W-Saltzman, line 93: It looks like you are commenting out unused code here.  This is what git is for.  It's ok to comment out the code while you are working.  remove it when finished.  You still have the old code in git.

Paul-W-Saltzman (approved):
Great job.  You used some advanced coding here.   I found a bug in the goal menu when you exit before choosing a single goal.  I left comments on the code.  But that's one of the challenges and not a requirement.  You where comment heavy and you left some commented code in the program.  I left notes about that, but you will find your balance. You met all the requirements and more.  I really liked your use of Spectre Console showing the progress.  Great Work! Keep Coding!

CodingTracker.A-Altemara/CodingTracker/ICodingTrackerDb.cs
A-Altemara: The XML comments enable you to hover over the method name and get info on what the function does and returns without going to the actual method. It makes reading through it easier. I've fixed the error in the view goal return.

CodingTracker.A-Altemara/CodingTracker/CodingDb.cs
A-Altemara: I don't know why it looks commented out, it is not. That code is used when a new DB is created and it populates the CodingTrackerTable with sample data.

CodingTracker.A-Altemara/CodingTracker/ICodingTrackerDb.cs
A-Altemara: Or am I misunderstanding the comments your referring to?

CodingTracker.A-Altemara/CodingTracker/CodingDb.cs
Paul-W-Saltzman: You did a great job on this program, and I already passed it.  in this comment I'm referring to lines 94 through 102 and other areas with comments.  Not the whole file.   Looking at it again I can see it is not code (sorry, my mistake).   It just has informational tags (for example <remarks> </remarks>) and comments, but it is commented out.  You could remove that whole section, and it wouldn't make a difference to the program.   As you move forward just be aware of how many comment and what comments are needed.  Most of your code is good enough that the comments are not needed.  It's actually easier to read without the comments getting in the way :).

Paul-W-Saltzman: Git hub removed my example because it looked like code haha, but the tags I'm referring to are in lines 96 and 98.

PR #142, opened 2024-10-05
hasona23-CodingTracker

hasona23:
Dear sir,
this is the coding tracker the final requirment for the olive green belt


arashi256 (changes requested):
Positive Aspects
I really enjoyed the convenience of being able to simply press "Enter" to default to today's date when adding a new session—very user-friendly!
The built-in help section is a great addition. It’s thoughtful to include that kind of guidance directly within the app, making it much easier for users to find what they need.

Suggestions for Improvement
I encountered an issue where the config.xml needed to be in the debug runtime folder to prevent a crash. You might consider reviewing the "Changing Your Working Directory" section from the habit tracker project, as it could provide useful insights for addressing this.

A key requirement is to ensure the application can handle all errors without crashing. Currently, certain scenarios lead to unresponsive behavior:
For instance, if an invalid date format (like "10/10/2024") is entered when trying to remove a session, the app hangs.
Also, if a correct date is entered but the description is left blank (just pressing "Enter"), it causes the application to hang as well.
When selecting a session to update, the UI could be more informative. Right now, it only shows the search terms, making it unclear whether the correct record has been retrieved.

Is searching by date the only option? Perhaps listing the last 5 most recent sessions? 
To enhance consistency, you could format the update results similarly to how you list sessions in the session history section. This could make it easier for users to locate and identify the right details.
Other Observations
It’s a bit unclear why the Database and CodingSession components are part of a shared library, as CodingSession seems to contain methods that may not be closely related. Streamlining the purpose of these classes could make your codebase easier to maintain and enhance readability.

Please review the Input validation especially.  This should always enforce correct input or give the user some guidance on what they are doing wrong.

hasona23:
Hi, thanks for your review @arashi256 . I apologise for the poor feedback of input the app didnt hang but I made so that it doesnt proceed until you make the correct input anyway i think this issue is fixed now and a clearer message is printed regarding the problem with use input
thanks.

arashi256 (changes requested):
- "It appears that format is incorrect pls enter correct format d-M-yyyy or press enter to skip" Pressing enter to skip doesn't seem to skip. 
- Please don't use cellphone-style abbreviations "pls" instead of "please", it looks unprofessional.
- When updating a session, the "old session" display only displays what you entered still, not the actual old session record data. It's still unclear if you have retrieved the correct entry.
- Update updates 0 records and doesn't seem to work.
- I couldn't get it to delete the record either. "Deleted 0 records"
- After this, when trying to add another session, it prints the date twice and seems to do nothing else after that yet still wants a time.
- With only one date field and 2 time fields, what happens if you have a coding session that starts at 11pm and ends at 2am? Will it calculate the duration correctly?

ct2


Sorry, buddy :( Please try again.

hasona23:
ok no problem I'll work on it now

hasona23:
Hi @arashi256  I resolved the issue you pointed out 
Details:
1. since I was using  the same function but without allowing to press enter to skip feature I solved this and made the message better and removed the press enter to skip for this part (Deletion and updating) . I think it would have been a better approach to separate the message and Input functionalities to allow more reusability
2. sorry for using cell-phone style messages I changed it
3. The updated 0 updates isnt a malfunction actually it is used to indicate if you entered the correct details of session. after all I solved this for delete and update and if it doesnt find any entries with details you described it will just show a clear message saying this 
4. the printing twice was mainly for me to make sure what i entered it what i get I believe this should be used primarily for debugging so I fixed it 
5. I prevented at the begining making end time  and starttime in diffrent dates and set a limit to the duration anyway i removed this and now it accepts having the start time at 23:00 and end 2:00 for example and the duration will be displayed normally as 3 hours and to fix this i just checked if the endtime hour is less than the start time and if that is the case i add a day to end date

arashi256 (approved):
All requirements satisfied. I'm going to approve this but please consider your UI flow. For example, you need to know the date of the session and something of the description from memory in order to update or delete a session. I found myself constantly having to enter nonsense data to get out of the current update or delete operation and listing all of them from the main screen and noting the session down on paper to enter the right details to update or delete becuase I couldn't remember them. This adds unnecessary friction to the application.

PR #152, opened 2024-11-01
CodingTracker

TwilightSaw:


arashi256 (changes requested):
Hi, 
Very nice! I like all the functionality. However, I was able to crash the application by selecting reports when there are no coding sessions: -
shot
Please patch, check over it again and resubmit and I will be happy to re-review! 👍

arashi256 (changes requested):
"appsettings.json" not found :) I thought it might be because the runtime directory is different to the current directory, but it is literally not in the project. 

shot

arashi256 (changes requested):
Hi, looks like this still needs some work, I'm afraid:-
- If I enter a coding session for a previous day (04.11.2024 - 10:00:00 - 14:00:00) and then try and delete the session by entering the date of the session, it cannot find it to delete it. 
- You don't tell the user what format to use prior to entering a date so you have to input a wrong one to find out what the actual format is. This is a listed requirement.
- Updating a session looks like it finds the right session to update (but doesn't display it so hard to tell), but attempting to modify the start or end times just give you "empty date".
- Same for delete functionality.

Room for improvement/things to think about but not necessary to get submission approval:
- If I enter a coding session for example 04.11.2024 starting at 11pm (23:00:00) and try and finish at say 2am (02:00:00) it says it's bad session input, presumably because it cannot cope with coding sessions that cross midnight.
- Keep your date formats consistent - you're asking for input as dd.MM.yyyy but you're displaying listed sessions as dd/MM/yyyy, pick one and stick to it. 
- Do you *really* need to account for seconds when entering a time?

TwilightSaw:
Some features can work incorrectly with US DateTime format

TheCSharpAcademy (changes requested):
@TwilightSaw Thanks for making the changes. 😄You're super close! There's just one issue that I can't reconcile with. ☹️Why are you trying to Parse all of the dates into a ToLongtime and then parsing into Duration? 

This is all over your code and I don't think it's necessary. To parse your times you should always use TryParse passing the specific format you expect. This is the best way to handle unexpected errors.

Usually we're not too strict in these initial projects but I'll request this change because using TryParse will avoid all sorts of headaches in a professional environment. You can find working examples of TryParse being used in other student's PR's, but let us know if you get stuck.😄

TheCSharpAcademy:
@TwilightSaw Thanks for the changes. ✅ I'm going to approve because I don't want you to be stuck, but the duration isn't being calculated properly. I'm not sure why  there's so much parsing in your code. You get a timespan by simply subtracting end time from start time: 
image

🏅🏅But you got your Olive Green Belt! 

📚I also recommend you go through our [Object Oriented Programming course](https://www.thecsharpacademy.com/course/1/article/1/500000/false) to increase your ability to write cleaner and more maintainable code.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁

PR #156, opened 2024-11-25
Submission

yemiOdetola:
Hi there,

Please take a look at my project when you have the time.
Any suggestion or feedback will be greatly appreciated 
🙏🏾


hasona23:
Hi @yemiOdetola . Good job doing the coding tracker project.   However unfortunately I can't accept this request.                              Fist of all you missed 2 of the requirements. <br/> 
1. one of the requirements was  to make a README file like in the habit logger which seems to be missing  
2. You didn't use Dapper ORM in you project but only depended on ADO.NET.

 also there are other things I would like to point out. 
1. in the update function and insert when entered date incorrectly the option of "00" for current date becomes unavailable. perhaps because it is not in the while loop. (check image)
image
 2. make db name more meaningful for instance instead of db.db make it CodingSession or coding_session.db
 3. also i prefer make the endtime dynamic so if i made a session for date 12-12-2012 then the endtime will be 12-12 or 13-12 and if i made it in 12-12 but with 02:00 and start in 18:00 then it automatically increase the day by 1 to 13-12 and that way i wont have a negative number duration
 4. Make the update paramters optional so the user wont have to seach for his record to get the parameters he doesnt want to use
 5. Use spectre.console to do things more conveniently like getting user input with Ansiconsole.Prompt which would make it easier and cleaner to get input and you can specify weather you want input to be null or not  and also display table in a good looking table
 6. Lastly use console.clear when done with an operation like adding a session to make console clean and easy to use and wait for user to press enter to remove the data of sessions after show all

Advices.
1- Dont make the dbquery static instead make an instance of it in the controller to be easily disposed. (Something for the future projects)
2- make the connection string readonly or in this case use const keyword to make it static and prevent it from having its value altered.
thanks.

yemiOdetola:
Thank you @hasona23 , I'll make the changes and revert!

hasona23:
Hi @yemiOdetola . Good job doing the changes. The project is accepted but if you don't mind here are some tips i would like you learn about for upcoming projects.
1- Dependency injection which could have been done in the controller instead of making an instance for every operation
2- Please take a look at spectre console website https://spectreconsole.net/ it would make it easier for making input especially for selecting the record to delete or update using SelectionPrompt.
3- Make sure console is always clear for the user and clean so he doesnt feel that the cli is filled with text that he needs to try figure out what he want (Like in debugging)

any way Congratulations on competing the project. 🎉 🎉 
 Good luck with next one . 😄

yemiOdetola:
Thanks for the insights, @hasona23 . Much appreciated!

PR #158, opened 2024-12-03
Project for review - CodingTracker 

DreamFXX:


DreamFXX:
IMPORTANT: For proper function of the App, add System.Configuration; at the beggining of the DatabaseManager.cs class.

PR #170, opened 2024-12-30
met base requirements

joshluca98:


TheCSharpAcademy (changes requested):
@joshluca98 Thanks for submitting! 😄 It's looking good except for one issue. 😟Please make sure I'm not allowed to input an end date that comes before a start date. You'll also need to handle scenarios where a shift starts in one day and finishes in the next day. 

image
 
Also, I noticed that you are validating the dates correctly, but it would be good to warn the user when they try to input an incorrect date/time. At the moment I just get prompted to input again.

joshluca98:
> @joshluca98 Thanks for submitting! 😄 It's looking good except for one issue. 😟Please make sure I'm not allowed to input an end date that comes before a start date. You'll also need to handle scenarios where a shift starts in one day and finishes in the next day.
> 
> Also, I noticed that you are validating the dates correctly, but it would be good to warn the user when they try to input an incorrect date/time. At the moment I just get prompted to input again.

Thanks for bringing those issues to my attention, they have been resolved. 

For the issue regarding end time being less than start time and still being accepted, I made it so that the user is notified of this (in case was typed by accident) and allows the user to either proceed (in which case the code assumes the user worked into the next day at the specified time and calculates the proper duration) or abort and return to the menu.

I have also added proper error messages when dates or times are entered in invalid format.

TheCSharpAcademy:
@joshluca98 Project approved! 😄✅ It's a shame you still haven't completed the challenges. Hopefully you'll get back to them eventually. 

🗺️And well done! You got your 🟩🟩 Olive Green Belt!

🧑🏻‍🏫The code is looking good. Just a small (optional) recommendation. Use the new namespaces to remove one level of indentation from your files: 
image

✋🏻The next project will be a step up in terms of complexity, but we recommend you do the challenge to learn about the "Pivot" keyword, a very important skill in your SQL toolbox.🫡

PR #178, opened 2025-01-27
Project Completed!

cacheMe512:
Project completed and ready for review. Many thanks!


hasona23 (approved):
Hi. sorry for the delay. I actually dont have any comments regarding the project.
everything is working and organised and readable. the only minor issue is when ending a session that is not started or starting a session that is not ended it show me the error as it closes quickly so you just need to add like a Console.Readline or something and press enter to continue.

keep up the good work 🎉

PR #184, opened 2025-02-11
Ready for PR

pkrzysiekk:


Dejmenek:
Hey @pkrzysiekk 👋,
Thanks for submitting😁. It's looking good except for one issue 🔴. I've also added some suggestions marked as 🟡.
___

🔴 When there are no coding sessions in the database and the program asks to select coding session, pressing `Enter` causes an exception because there is nothing to select. It would be better to validate this by not prompting the user for a session when none are available, and inform them that they need to add a session first. The same applies when deleting a coding session.
image

___

🟡 When updating a coding session, it would be good to know what I'm updating. The user won't have to remember what was previously entered. You can show the record being updated.

🟡 Right now, the console output stacks up after each action, It would be better to clear the console after each input or key action to keep the output clean and more readable. You can use `Console.Clear()` to refresh the screen before displaying new prompts.
image

🟡 I've noticed a lot of empty lines in your files. Next time make sure to clean up a bit before submitting as we tend to get more strict with code tidiness👀. You can read our code conventions [here.](https://www.thecsharpacademy.com/article/58/code-conventions)

___

🟢 Well done for completing challenge with using Stopwatch so the user can track the session as it happens🤗

CodingTracker.pkrzysiekk/CodingTracker/Controllers/UserInput.cs
```diff
@@ -0,0 +1,104 @@
...
+            sw.Start();
+            AnsiConsole.Prompt(new TextPrompt<string>("[violet] Type anything to stop the session and save the elapsed time[/]"));
+            sw.Stop();
+            return sw.Elapsed;
+        }
+        public static CodingSession GetUserRecordInput()
+        {
+            DateTime startDate;
+            DateTime endDate;
+            bool manualSession = AnsiConsole.Prompt(new TextPrompt<bool>("Do you want to start the session now and end when you're finished?")
```
Dejmenek: 🟡 You can simplify it by using `AnsiConsole.Confirm("Do you want to start the session now and end when you're finished?")`.

CodingTracker.pkrzysiekk/CodingTracker/Controllers/UserInput.cs
```diff
@@ -0,0 +1,104 @@
...
+        public static int CalculateDateDifference(DateTime startDate, DateTime endDate)
+        {
+            var dateDiff= endDate - startDate;
+            return (int)dateDiff.TotalMinutes; 
+        }
+        public static TimeSpan GetUserTime()
+        {
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+            AnsiConsole.Prompt(new TextPrompt<string>("[violet] Type anything to stop the session and save the elapsed time[/]"));
```
Dejmenek: 🟡 You can simplify it by using `AnsiConsole.Ask<string>("[violet]Type anything to stop the session and save the time:[/]")`

pkrzysiekk:
Thank you for your feedback! I think I fixed all the issues.

Dejmenek:
@pkrzysiekk Project approved✔
Looking forward to seeing more great work from you in future projects!😎

PR #186, opened 2025-02-23
Finished project

mrgee1978:
Please critique my code and help me improve. A little disappointed that I did not feel I was able to do the challenges yet. But I will keep learning and improving.


TheCSharpAcademy (changes requested):
@mrgee1978 Thanks for submitting and apologies for the delay! Unfortunately I ran into problems when trying to add a record. You should use TryParse or TryParseExact (recommended) instead of parse, otherwise computers with a different date format will get this error: 
image
 
Could you please fix it so I can approve? 😄

mrgee1978:
@TheCSharpAcademy thank you so much for reviewing my code. On the method that threw the exception, I had already done a check to make sure that the date string passed to the method was a valid one in the GetValidDateString() method. I have been trying to recreate the problem by entering the date string that appears in your screen capture and it seems to work fine for me I don't get the exception, so I was wondering if there was anyway that you could tell me everything that you entered when trying to add a coding session. I want to understand exactly what I did wrong to cause it to throw an exception and this would help me tremendously so that I can refactor the code. Thank you so very much for all of your help.

nwdorian (changes requested):
Hey @mrgee1978 👋,

sorry for the long reply, it looks like this got lost in our backlog.

---

I looked over the issue with `DateTime` parsing throwing an exception. It seems to be caused by `GetValidStartDate` method, because it's using `DateTime.Parse` which uses the **machine’s current culture** by default, which may not match `InvariantCulture`, and therefore **can throw a `FormatException`**, even though it passed the earlier check.

A fix would be to use `DateTime.ParseExact(validDateString, dateFormat, CultureInfo.InvariantCulture)`.

But at that point you might as well use `DateTime.TryParseExact()` like in the previous method.

---

If you end up fixing the bug, push the changes to your fork. The PR will get updated and I will be notified to check. ☺️

Best regards,  
@nwdorian

PR #192, opened 2025-03-12
J-Mels Coding Tracker

J-Mels:


Dejmenek (approved):
Hey @J-Mels 👋,
I've reviewed your project and it looks great👌. I noticed that you haven't done any of the challenges😟. While they're not required, I encourage you to dive into them. They're a great opportunity to strengthen your skills💪.

___

🟢 You've fulfilled all of the project requirements🎉

🟡 For future projects, please organize your submission in a folder following this format: `[ProjectName].[GitHubUsername]` (e.g., `CodingTracker.J-Mels`), in accordance with the academy's [code review guidelines](https://www.thecsharpacademy.com/article/57/review-guidelines).

🟡 I've seen that you've included some comments in your code. While comments can be helpful, they're not necessary if the code is already self-explanatory. It's best to use them only when providing context or explaining something that the code itself cannot clearly convey.

🟡 Your program throws an exception when the user provides an end time before the start time. Since this is an expected user input error rather than an exceptional case, it might be better to handle it with validation instead. A more user-friendly approach would be to display an error message and prompt the user to enter a valid time range instead of terminating the program.

___

I'll mark your project as approved✔. Keep up the good work😁

PR #193, opened 2025-03-12
Project Submission

Andrewki44:


hasona23 (changes requested):
Hi @Andrewki44 
Well done the project is really nice and i liked the design of console however there are a few things i must point out # Critical
    - You didnt add values to paramters when saving sessions wether for update or add 
    this is the error i see which happens becuase the variable name must be same as the paramter so either
    ```
    conn.Execute(commandText, new {sessionStart=session.SessionStart,sessionEnd=session.SessionEnd,seconds=session.Seconds});
    ```
    or

    ```
     string commandText = @"
                INSERT INTO CodingSession (Start, End, Duration)
                VALUES (@SessionStart, @SessionEnd, @Seconds)
            ;";

    ```
    I just changed the paramter names so be same as variables in the coding session (basically uppper case first letter)

# Suggestions
1- Validate dates. Dont allow session when entered manually to have Zero duration or negative duration
2- When seeding data into DB you must make them varied different dates and durations
3- Use try-catch statements when doing DB functions to avoid program crashing instead of make try-catch statement on the whole operation in the program.cs file
4- Goto statements arent recommended to use much since they make program harder to read and i dont see why use them since you are already
in a while loop which will keep running until it meets a certain condition and in this case make a bool called isRunning 
and when exit change it  to false and get out of loop
5- Make code modular. I liked what you did with menu and validation but would have been even better if you in program.cs put each
case in a function you dont have to make a new class but make it more readable instead of having to read a while loop with a switch case 
and nested if conditions all at once instead make the loop and switch have functions like 
```
do{
    switch(myCase){
        case 1:
            AddSession();
        case 2:
            Remove();
        case..... etc
    }
}while(isRunning)
```
this will also make it easier for you to add new features and expand on the project.

6- Use enums or even constants for the menu options this makes it easier to do operations and identitfy them instead of using raw string
if you dont want enums you can use const string variables like 
``` 
    const string AddSessionOption = "Add session"
```


anyway just by doing the critical ones you will be good to go

TheCSharpAcademy:
@Andrewki44  It's been two weeks since the suggested changes, so we're archiving this project. Feel free to reopen the PR and resubmit in the academy when you're ready. 🤓@hasona23 You'll get the points regardless.

PR #200, opened 2025-03-31
Project finished!

Lepros311:
Thank you for your review of this project!


TheCSharpAcademy (changes requested):
@Lepros311 Hello! Sorry it took so long to review. It's looking great, I just found one small bug. At the moment I'm able to add an end time that comes before that start time. That shouldn't be allowed. Check out the last record:

image

Could you please fix so I can approve? It should be a good bug fixing exercise 🫡

Lepros311:
Thanks. I just updated the code and pushed it again.

-Andrew (Lepros)

TheCSharpAcademy:
@Lepros311 I'm approving and closing since we've discussed this project at length 😊

PR #202, opened 2025-04-09
Initial Merge into Base

geicoandsonic:
Finished the project! I learned a ton about integrating SQL into C# applications, as well as using NuGet packages.


thomas-henley:
Hi @geicoandsonic 👋 

I'm going to take a look at your submission for the Coding Tracker. A lot of my comments will just be bits of advice to consider, not necessarily something you need to change. I'll annotate them for clarity:

🛑 : Something that needs to be fixed before the merge request can be approved.
🟨 : No fix needed, just something to think about. These will usually be coding style notes.
💚 : Something worth calling attention to for being particularly well done.

I'll comment here again when my review is complete.

CodingTracker/App.config
thomas-henley, line 1: 🟨 : `App.config` is traditionally used in .NET Framework, while more modern .NET Core applications favor `appsettings.json` for configuration. Consider using `appsettings.json` for future projects.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
...
+
+                SqliteDataReader reader = tableCmd.ExecuteReader();
+
+                if (!reader.HasRows)
+                {
+                    Console.WriteLine($"\nNo record found with the given ID {id}");
+                    connection.Close(); //Close database
+                    return false;
+                }
+                else
```
thomas-henley, line 139: 🟨 : You could remove this `else` to reduce the nesting in this method. The `if (!reader.HasRows)` block above exits the method, so it works as a [guard clause](https://en.wikipedia.org/wiki/Guard_(computer_science)) and does not need an `else`.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
+﻿using CodingTracker.UILayer;
+using Microsoft.Data.Sqlite;
+namespace CodingTracker
+{
+    public static class DatabaseManager
+    {
+        private static string _connectionString = "";
+        private static string _databaseName = "";
+        private static DateTime startTime;
```
thomas-henley, line 9: 🟨 : Add a blank line before and after methods.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
...
+                tableCmd.CommandText =
+                    @$"CREATE TABLE IF NOT EXISTS {_databaseName} (
+                        Id INTEGER PRIMARY KEY AUTOINCREMENT,
+                        StartTime TEXT,
+                        EndTime TEXT,
+                        Duration INTEGER)";
+
+                tableCmd.ExecuteNonQuery();
+            }
+        }
```
thomas-henley, line 28: 🟨 : Add a blank line before and after methods.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
...
+                }
+            }
+            return time;
+        }
+
+        public static void AddRecord()
+        {
+            Console.Clear();
+            Display.PrintHeader("Add Record");
+            string startTime, endTime = "";
```
thomas-henley, line 48: 🟨 : Avoid ambiguous variable declarations like this. Here `endTime` is being initialized to `""`, but `startTime` is _not_ and will instead be `null`. It's not necessarily wrong, if that's the result you intended, but the lack of clarity is reason enough to avoid it. Better to declare variables in separate statements completely.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
...
+                }
+                Console.WriteLine("\nPlease enter the time you finished coding.");
+                endTime = GetTime();
+            }
+            Console.Clear();
+            Console.WriteLine($"Entry added successfully.");
+            
+            using (var connection = new SqliteConnection(_connectionString))
+            {
+                connection.Open(); //Open database
```
thomas-henley, line 67: 🟨 : No need to add comments like this when it's very clear what the code is doing. Check out [the C# Academy style guide](https://www.thecsharpacademy.com/article/58/code-conventions) for a little more detail.

CodingTracker/UILayer/Display.cs
```diff
@@ -0,0 +1,45 @@
...
+                .AddChoices(new[]
+                {
+                    "Close Application", "View All Records", "Add Record", "Update Existing Record", "Delete Record", "Delete All Records", "Timer Status", "Stop Timer"
+                }));
+            }
+
+            return menuChoice;
+        }
+
+        public static void PrintHeader(string headerInput, string color = "blue")
```
thomas-henley, line 37: 💚 : Very nice flexible method and looks good in the running app.

CodingTracker/DatabaseManager.cs
```diff
@@ -0,0 +1,206 @@
...
+                    connection.Close(); //Close database
+                    
+                    return true;
+                }               
+            }
+        }
+
+        public static void StartTimer()
+        {
+            startTime = DateTime.Now;
```
thomas-henley, line 177: 🟨 : The variable name `startTime` as a static class variable gets [shadowed](https://en.wikipedia.org/wiki/Variable_shadowing) by local variables with the same name within your methods (like on line 144, for example). It doesn't break anything in your app but in general you want to avoid it -- it can lead to confusing and difficult-to-track bugs. When two variables can be accessed within one scope, they should almost always have different names.

CodingTracker/DatabaseValidator.cs
```diff
@@ -0,0 +1,48 @@
...
+                return true;
+            }
+            else
+            {
+                Console.WriteLine("Invalid date format. Please enter the date in the format 'MM-dd-yyyy HH:mm:ss'.");
+                return false;
+            }
+        }
+
+        //If the start time is before the end we have a problem
```
thomas-henley, line 20: 💚 : Good catch. This is an easy validation to overlook. Although I think you meant this comment the other way around 😄 .

CodingTracker/DatabaseValidator.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if(DateTime.Parse(startTime).Subtract(DateTime.Parse(endTime)) > TimeSpan.Zero)
+            {
+                Console.WriteLine("Start time must be before end time.");
+                return false;
+            }
+            return true;
+        }
+
+        public static int GetQuantity(string message)
```
thomas-henley, line 31: 🟨 : This method probably doesn't belong here. You are combining the user input and the validation into a single method, instead of collecting input (probably with your `UserInputManager`) and then validating it here.

thomas-henley:
Great job @geicoandsonic !

I know it looks like a lot of comments, but most of them are just minor things to think about. Feel free to click "Resolve conversation" on any of them, or ask questions in the thread if something is unclear.

However, there are [two requirements for the assignment](https://www.thecsharpacademy.com/project/13/coding-tracker) that were not met:

🛑 : `The user shouldn't input the duration of the session. It should be calculated based on the Start and End times, in a separate "CalculateDuration" method.`

I'm not concerned about this one. The important part of this requirement is that the user does not manually enter the duration. But I will point out that you repeat the calculation several times, so it would be best practice to extract that to a method in the `CodingSession` class.

🛑 : `You need to use Dapper ORM for the data access instead of ADO.NET. (This requirement was included in Feb/2024)`

Is there any particular reason you did not use Dapper?

TheCSharpAcademy:
@geicoandsonic  It's been two weeks since the suggested changes, so we're archiving this project. Feel free to reopen the PR and resubmit in the academy when you're ready. 🤓@thomas-henley You'll get the points regardless.

PR #212, opened 2025-06-17
Project finished!

santisica29:
Please let me know what I can improve for my next project and what I am doing well.

I appreciate feedback on my use of Separation of Concerns.


santisica29.CodingTracker/CodingTracker/Data/DatabaseInitializer.cs
```diff
@@ -1,5 +1,4 @@
 ﻿using Microsoft.Data.Sqlite;
-using System.Configuration;
```
nwdorian: 🔴 Project doesn't build

🐞 Removing `using System.Configuration;` causes errors so I can't run the project locally. Codacy can have false-positives with certain code, if it suggests to delete something that is required for your project to work, you can ignore it and leave a note for reviewers.

💡A good rule of thumb is to always build and run your code before committing it to source control. This way you will make sure to have a backup of a working copy.

santisica29.CodingTracker/CodingTracker/Controller/BaseController.cs
nwdorian, line 1: 🟡 **Base class**

💡Nice to see that you tried to implement inheritance by using an abstract class. I do feel like these methods belong elsewhere, probably in `Helpers.cs`, since the `CodingController.cs` already does lots of other things apart from just displaying prompts.

santisica29.CodingTracker/CodingTracker/Controller/CodingController.cs
nwdorian, line 1: 🔴**Separation of concerns**

➡️ Extract the code for database interaction with SQL into a separate class - I suggest adding it into `Data` folder in a class called `DatabaseMethods.cs` or something similar. Then you can call these methods from the controller and reduce some lines of code. This approach also helps with reusability.

santisica29.CodingTracker/CodingTracker/Enums.cs
nwdorian, line 1: 🟢 **Enums**

⭐ Good job on using enums instead of magic strings. Look into the `[DisplayName]` attribute and use a converter with Spectre.Console to show a human friendly version. e.g. `Add Coding Session`.

santisica29.CodingTracker/CodingTracker/Helpers.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    internal static bool IsFormattedCorrectly(string date, string format)
+    {
+        if (!DateTime.TryParseExact(date, "yyyy-MM-dd HH:mm", new CultureInfo("en-US"), DateTimeStyles.None, out _))
+        {
+            return false;
+        }
+
+        return true;
+    }
```
nwdorian: 🔴 **Missing Requirement**

🐞  As per this projects [requirements](https://thecsharpacademy.com/project/13/coding-tracker), you are required to have your validation logic in a separate class `Validation.cs` (you will thank yourself when you get to the Unit Testing project).

santisica29.CodingTracker/CodingTracker/coding_tracker.db
nwdorian, line 1: 🟠 **Database in Repo**

💡 You shouldn't be committing your database file to source control. Add it to the `.gitignore`.

nwdorian (changes requested):
Hey @santisica29👋,

Great work on your **Coding Tracker** project submission! 🎉

I have performed a code review. Review/ignore any comments as you wish.

---

Please take care of issues marked with 🔴, as they block my approval. 🚫

Feel free to reach out if you have any questions or if you'd like to collaborate on any of these points.

---

Please edit your code to fix these issues then push the changes to your fork. The code will get updated in the PR and I will be notified to check again. ☺️

Best regards,  
@nwdorian

santisica29.CodingTracker/CodingTracker/Controller/CodingController.cs
santisica29: Are you speaking about the GetReport and GetReportOfTotalAndAvg or more lines of code? I don't understand completely how to separate them from the codingController class could you give me an example?

santisica29.CodingTracker/CodingTracker/Enums.cs
santisica29: Thanks for the info I ended up using [Description] instead since it look easier to understand and apply. 

With the help of chatgpt end up creating a method that gets the description attribute of a enum and then i've used UseConverter to display that text instead of the default enum.ToString(). 

I think I've already used the UseConverter in a similar manner (showing the names of the author and name of the book in a previous project) but i forgot about it, it was a good refresher thanks!

santisica29.CodingTracker/CodingTracker/Controller/CodingController.cs
nwdorian: For example `AddSession` method

after you get the input from user and create a new coding session I'd like to see a method call like this

```
var affectedRows = databaseMethods.CreateSession(session);
```

this method would be extracted into `DatabaseMethods.cs` class and would probably look something like this

```
public int CreateSession(CodingSession newSession)
{
    using var connection = new SqliteConnection(DatabaseInitializer.GetConnectionString());
    var sql =
            @$"INSERT INTO {DatabaseInitializer.GetDBName()} (startTime, endTime, duration)
               VALUES (@StartTime, @EndTime, @Duration)";

    var affectedRows = connection.Execute(sql, new { StartTime = newSession.StartTime, EndTime = newSession.EndTime, Duration = newSession.Duration.ToString() });
    
    return affectedRows;
}
```

To use it in the controller you would just create a new instance of the `DatabaseMethods.cs` class:
```
internal class CodingController
{
    private readonly DatabaseMethods databaseMethods = new();
    
    // ... rest of the code
```

nwdorian: @santisica29 

I update the code in my example as I noticed some mistakes (sorry, it's hard to write without IDE support 😄 )

Let me know when you are done with all changes.

santisica29: Oh ok I think I get it now, I should move the parts that comunicate with the database to the DatabaseMethod class to keep the separation of concerns and then just call those methods from my CodingController class. Let me see if I can do it. Thanks for the example it's very clear!

santisica29: I have a quick question. What are the advantages and disadvantages of keeping the new DatabaseMethods class static, as opposed to creating an instance of it and using that? I made my DatabaseInitializer class static so I could call its methods directly, but I'm unsure whether this is considered a bad practice.

nwdorian: In general we should not use static methods at all because they don't align with OOP principles. With that said, static classes and methods certainly have their place in the language, for example when defining extension methods. The way I personally approach it, is to make nothing static and only if the compiler gives me a hint that says something like "you can make this static" I consider it.

nwdorian: While we are on this refactoring topic, make sure to also move all other methods that have SQL in them to the `DatabaseMethods` class. There's a few of them at the bottom of the controller class that only interact with database so it would make sense to have them together with the ones you will exctract in the new class.

santisica29: Perfect, on it!

nwdorian:
Hey @santisica29 👋,

any further progress on this? 

You would still have to extract the validation logic to `Validation.cs` class to have the project approved.

santisica29:
Hey @nwdorian apologies for taking so long. 

I moved all the sql logic to the DatabaseMethods.cs and the validation logic to a new Validation class.

Also I've created a .gitignore folder where I put the database file to prevent people from accessing my data.

Let me know If these changes were what you had in mind or if I have to correct something else.

Thanks again for taking the time to review my code, it's much appreciated!

santisica29.CodingTracker/CodingTracker/CodingTracker.csproj
```diff
@@ -0,0 +1,28 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <COMReference Include="{bee4bfec-6683-3e67-9167-3c0cbc68f40a}">
+      <WrapperTool>tlbimp</WrapperTool>
+      <VersionMinor>4</VersionMinor>
+      <VersionMajor>2</VersionMajor>
+      <Guid>bee4bfec-6683-3e67-9167-3c0cbc68f40a</Guid>
+    </COMReference>
+  </ItemGroup>
```
nwdorian, line 17: 🟠 **Build fails**

This `<ItemGroup>` fails the build on my machine with the error below. By looking it up it seems to have something to do with .NET Framework (so the old windows-only .NET version). I'm not quite sure why this is here, double check this on the next project submission.

image

santisica29.CodingTracker/CodingTracker/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "CodingTracker": {
+      "commandName": "Project",
+      "workingDirectory": "C:\\Users\\Santi\\Desktop\\c#\\C#AcademyForks\\CodeReviews.Console.CodingTracker\\santisica29.CodingTracker\\CodingTracker"
+    }
+  }
+}
```
nwdorian, line 8: 🟠 **Instructions for running the app**

This also prevented the app from running, because this folder structure obviously don't exist on my machine. 

A very very useful and often overlooked section of README is installation and running instructions. Include something like "change the working directory in launchSettings.json" along with any other steps required for application startup.

santisica29.CodingTracker/CodingTracker/Data/DatabaseMethods.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        using var connection = new SqliteConnection(DatabaseInitializer.GetConnectionString());
+
+        if (sql == null) sql = $"SELECT * FROM coding_tracker";
+
+        var listFromDB = connection.Query(sql).ToList();
+
+        if (listFromDB.Count == 0) return null;
+
+        var listOfCodingSessions = Helpers.ParseAnonObjToCodingSession(listFromDB);
```
nwdorian, line 59: 🟠 **Query to the type directly**

💡Dapper `Query<T>` is a generic method, which means you can pass in a type instead of the `<T>` parameter. By passing the `CodingSession` type you could complete avoid the `ParseAnonObjToCodingSession` method.

🟠 **Avoid returning null collections**

💡I think you are overusing null just a bit in your code. You will quite soon start to run into null problems and you will spend more time handling null exceptions then doing actual coding 😄 Instead return an empty list with `Enumerable.Empty<CodingSession>().ToList()`.

Here is the improved code with above suggestions applied:

```suggestion
        var listFromDB = connection.Query<CodingSession>(sql).ToList();

        if (listFromDB.Count == 0) return Enumerable.Empty<CodingSession>().ToList();
```

santisica29.CodingTracker/CodingTracker/Controller/CodingController.cs
```diff
@@ -0,0 +1,161 @@
...
+        var endTime = timer;
+
+        AddSession(startTime, endTime);
+    }
+
+    public void ViewSessions(List<CodingSession>? list = null, List<string> additionalList = null)
+    {
+        Validation.CheckIfListIsNullOrEmpty(list);
+
+        Helpers.CreateTable(list, ["ID", "Start Time", "End Time", "Duration"]);
```
nwdorian, line 126: 🟠 **Null exception**

💡If you go to "View Report of Coding Session" or "View Coding Session" with an empty database there will be a null exception which closes the app. This is what I was saying earlier with overusing null, try and find ways to avoid using it - it's not **always** possible, but it's also not good relying on it as much. 

➡️ Your `CheckIfListIsNullOrEmpty(list)` should technically not pass the code further but the `return` is inside it
so it won't actually stop the execution of the current method - I believe you misinterpreted that part. Something like this would prevent the execuction and reaching the null reference exception from `CreateTable` method:
```suggestion
        if (Validation.CheckIfListIsNullOrEmpty(list))
        {
            AnsiConsole.MarkupLine("[red]No data found.[/]");
            AnsiConsole.MarkupLine("Press Any Key to Continue.");
            Console.ReadKey();
            return;
        }

        Helpers.CreateTable(list, ["ID", "Start Time", "End Time", "Duration"]);
```

santisica29.CodingTracker/CodingTracker/Helpers.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Diagnostics;
+using System.Globalization;
+using System.Reflection;
+
+namespace CodingTracker;
+internal static class Helpers
+{
+    private static readonly UserInterface userInterface = new();
+
+    internal static List<CodingSession> ParseAnonObjToCodingSession(List<dynamic> listFromDB)
```
nwdorian, line 15: 🟠 **Using `dynamic` type**

💡I mentioned in a previous comment that you could use the `Query<CodingSession>` and completely avoid this parsing method. As a rule of thumb if you have to use `dynamic` or `object` somewhere - you're doing something wrong. Well, not neccessarily wrong, but it can definetly be done in a different way that's more aligned with strongly typed language paradigm.

santisica29.CodingTracker/CodingTracker/Validation.cs
```diff
@@ -0,0 +1,38 @@
...
+    internal static void CheckIfListIsNullOrEmpty<T>(List<T>? list)
+    {
+        if (list == null || list.Count == 0)
+        {
+            AnsiConsole.MarkupLine("[red]No data found.[/]");
+            AnsiConsole.MarkupLine("Press Any Key to Continue.");
+            Console.ReadKey();
+            return;
+        }
+    }
```
nwdorian, line 37: 🟠 **Return bool from validation methods**

👉 Related to `ViewSessions` method from the controller - if we return an empty list instead of null and return bool from validation methods, this is what the method would look like:
```suggestion
    internal static bool IsListEmpty<T>(List<T> list)
    {
        return list.Count == 0;
    }
```

santisica29.CodingTracker/CodingTracker/Data/DatabaseMethods.cs
```diff
@@ -0,0 +1,100 @@
...
+using Dapper;
+using Microsoft.Data.Sqlite;
+using System.Globalization;
+
+namespace CodingTracker.Data;
+
+internal class DatabaseMethods
+{
+    public int CreateSession(object newSession)
+    {
```
nwdorian, line 12: 🟠 **Using `object`**

👉 As mentioned before using `object` or `dynamic` is kind of a code smell. There is certainly a place for them in the language, but we want to take advantage of strong types - why are we not passing a type of `CodingSession` in here ?

nwdorian (approved):
Hey @santisica29👋,

Thank you for your effort and persistence while working on this project! 💪 

I've tested it locally and finished the code review. 

---

🟢 **Requirements**

---

🏅 **Extra challenges**

✅ Stopwatch feature
✅ Filtering coding sessions
✅ Create total and average reports

---

💭There are still some issues I've addressed in comments, but the application works very well overall, it's intuitive and easy to use. I like the stopwatch functionality a lot. Going to steal some of that implementation myself 😄 

➡️ We've had quite a bit of back and forth on this project and I'd like to see you advance further in the academy roadmap. You can take care of potential improvements if you have the time in the future.

santisica29.CodingTracker/CodingTracker/Properties/launchSettings.json
santisica29: I see how this would be an issue for others, is there a way to dynamically alter this so people don't have to change the working directory themselves?

santisica29.CodingTracker/CodingTracker/Data/DatabaseMethods.cs
santisica29: I used object because I passed the CodingSession object parsed into a object that uses strings since sqlite doesn't accept dates. Let me see If i can modify it

PR #213, opened 2025-07-07
Completed

SheheryarRaza:


hasona23:
Hi @SheheryarRaza.  How are you? Well done finishing the code tracker project
the project is fine yet I have a few suggestion.

1- Seed Data. Make like a DataSeeder class that puts fake data into the database for easier testing instead of having to write it yourself.
2- Duration is just a calculated variable so you don't have to make it as a variable and store it in Db instead could just do
```
public TimeSpan Duration => EndTime-StartTime
or
public TimeSpan Duration() => EndTime-StartTime
```

and that's it the code is really nice and well written.
Good luck on the next one  and Happy Coding

PR #215, opened 2025-07-11
Coding Tracker Project

Gincaza:
pretty fun project. I'm trying to use clean architecture

it was a little bit of overengineering but I really just wanted to put my hands-on good practice


BusinessLogicLayer/BLLClass.cs
```diff
@@ -0,0 +1,156 @@
+﻿using BusinessLogicLayer.ComunicationClasses;
+using System.Globalization;
+using DataClasses.BLLClasses;
+using DataClasses.ConfigurationClass;
+
+namespace BusinessLogicLayer
+{
```
nwdorian: 🟠 **Namespaces**  

❓As per the academy's [code-conventions](https://thecsharpacademy.com/article/58/code-conventions), use file-scoped namespaces.  

➡️ From C# 10 we are able to remove the namespace block and save one indentation level, which optimizes the use of space and improves readability.

💡 If you're using Visual Studio 2022, here's a short article to setup [file-scoped namespaces by default](https://darthpedro.net/2024/01/29/configure-visual-studio-to-use-file-scoped-namespaces-on-new-classes/).
```suggestion
namespace BusinessLogicLayer;
```

BusinessLogicLayer/BLLClass.cs
```diff
@@ -0,0 +1,156 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                return new OperationResult(false, ex.Message);
+            }
+        }
+
+        //calc the difference by time
+        private TimeSpan? TimeDifferenceCalc(string startDate, string endDate)
```
nwdorian: 🟡 **Code comments**

👉 It's fine to add comments while developing the application if it's going to help you with pseudo-code or to remember some specific logic. When you finish a project, always go back through the code for a polish pass and clean up unnecessary comments.  

💭 Remember, our code should be self-explanatory - with good naming and structure we shouldn't need any comments. This method name is clear enough that we don't need this comment. You shouldn't use abbreviations for variable names, a more suitable name would be `CalculateTimeDifference()`

BusinessLogicLayer/BusinessLogicLayer.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\DataAccessLayer\DataAccessLayer.csproj" />
```
nwdorian, line 10: 🟠 **Clean architecture dependencies**

👉 In CA all dependencies should point inward - currently you have an inner layer (BLL) depending on (referencing) the outer layer (DAL). 

➡️ You can see a dependency graph of your project in the image below. The reference in red is not good. The BLL and DAL should have switched places in the graph.

image

DataAccessLayer/DatabaseManager.cs
```diff
@@ -0,0 +1,123 @@
...
+            this.connectionString = connectionString;
+
+            InitializeDatabase();
+        }
+
+        private void InitializeDatabase()
+        {
+            using (var connection = new SqliteConnection(connectionString))
+            {
+                connection.Open();
```
nwdorian: 🟠 **Using blocks**

💡C# 8 introduced a [using declarations](https://learn.microsoft.com/hr-hr/dotnet/csharp/language-reference/proposals/csharp-8.0/using#using-declaration) feature. We are able to remove a level of indentation by utilizing `using` **declarations** instead of `using` blocks and improve readability.
```suggestion
            using var connection = new SqliteConnection(connectionString);
            connection.Open();
```

DataClasses/DataLayerClasses/CodingSessionEntity.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace DataClasses.DataLayerClasses
+{
+    public class CodingSessionEntity
+    {
+        public int Id { get; set; }
+        public required string StartDate { get; set; }
+        public required string EndDate { get; set; }
+        public required string AllTime { get; set; }
+    }
```
nwdorian: 🟡 **Naming**

👉 Maybe a more suitable name for this property would be just `Duration`.

nwdorian (changes requested):
Hey @Gincaza 👋,

Amazing work on your **Coding Tracker** project submission! 🎉

I have performed a code review and left some comments on things that I like and also on areas to improve. ☺️

Please take care of issues marked with 🔴, as they block my approval. 🚫

Feel free to reach out if you have any questions or if you'd like to collaborate on any of these points.

---

🟢 **Requirements**

---

🟠 **Solution Structure**

👉 As per the academy's [Code Reviews](https://thecsharpacademy.com/article/52/code-reviews) page, I was expecting your project submission to be contained in a folder with the format: **ProjectName.GitHubUserName**

➡️ For example, `CodingTracker.Gincaza` with all project related files and folders inside it.

💡 This will make it easier for code-reviewers to review your code in the future. It will also make sense when you get to the projects which contain other students code.

---

🟡 **UX improvements**

➡️ When entering date and time it's a bit cumbersome to also input seconds, having just the hour and minutes for input would be sufficient.
➡️ It would be nice to display Total Time in the table of all sessions in a more friendly format, e.g. like **1h 30m**

---
🟢 **Clean Architecture**

👉 I think it's really good you're trying to use Clean Architecture. It is probably an overkill for apps this simple, but I think what you did in this project is great intro to CA, creating all the projects and figuring out where classes go will help you learn best practices. 

💭 Related to the comment about references - since you are not using interfaces you will not be able to solve the project reference issue. But if you were to try and solve the reference issue with interfaces you will run into Dependency Injection pattern and if you're not familiar with it, maybe leave it for next projects. 

💡 I'm also wondering if you are reading a book on CA since your project names are very dogmatic. I believe any online article or video tutorial would go with more conventional names, so I'm genuinely curious if you are working through it alongside a book? A naming convention for CA projects would be `ApplicationName.LayerType` - here's what it would look like:

- `DataClasses` - `CodingTracker.Domain` (or Core)
- `BusinessLogicLayer` - `CodingTracker.Application` (or UseCases)
- `DataAccessLayer` - `CodingTracker.Infrastructure`
- `PresentationLayer` - `CodingTracker.Presentation` (it's also common for this layer to be named by the project type - in your case `CodingTracker.Console`, and for web applications `CodingTracker.WebApi` or just Web)

➡️ If you're interested, here are some resources on CA that were useful to me when I was learning about it
- [Ardalis CA](https://www.youtube.com/live/l1atr0trldU?si=Lj3GHk-iG3_0_mde) no code along, write down notes only
- [Codewrinkles CA](https://youtu.be/ZXGiactmoj0?si=uQ2Pn4N8cUCjuoXZ) code along with this one
- [Julio Casal CA](https://youtu.be/PPidZIx7GwU?si=hp8reNzTrLi-zpNG) no code along, no notes, enforce knowledge from first two and compare approaches

---

🔴 **DateTime Validation**

🐞 You currently have a bug where the validation for start and end times is missing. It's possible to input an end time that's before start time, putting data into a "broken" state.

👉 Introduce another validation rule to check if end time is after start time, then add it to relevant places in code.

image

--- 

Please edit your code to fix the 🔴 issues then push the changes to your fork. The code will get updated in the PR and I will be notified to check again. ☺️

Best regards,
@nwdorian

BusinessLogicLayer/BusinessLogicLayer.csproj
Gincaza: Thanks for the feedback! 😊

Regarding the dependency direction in Clean Architecture: I believe the current setup actually follows the correct structure. The BusinessLogic layer (BLL) depends on the DataAccess layer (DAL), not the other way around.

image

One idea I had was to introduce an intermediary class to handle the communication between BLL and DAL. This class would be responsible for parsing and translating data types between the layers, ensuring a cleaner separation of concerns. My intention was that any data transformation (parsing) would happen during function calls from BLL to DAL and in the return values from DAL to BLL.

image

Do you think introducing such an intermediary would improve the architecture further? I'd love to hear your thoughts or suggestions on that!

DataAccessLayer/DatabaseManager.cs
Gincaza: always good to learn new tricks :=)

BusinessLogicLayer/BusinessLogicLayer.csproj
nwdorian: @Gincaza

## Clean architecture

Now that is saw it, the structure in your diagram is N-Layer architecture. The dependencies flow towards the database which Clean Architecture solves by focusing on the domain instead of the database. 

If you watch any of the videos I recommended you would see that there is an emphasis on NOT having an inner layer reference outer layers.

**From Ardalis video:**

image

- You have a reference to infrastructure project which is a violation of CA rules.


**From Codewrinkles video:**

image

- You can see that the BLL (Application) layer only depends on the domain and doesn't reference any other project

**From Julio Casal video:**

image

- _"Nothing in an inner circle knows about anything in an outer circle"_
- Your BLL (Use Cases or Application) layer knows about the database layer through the reference. The dependencies flow outwards which is again, a violation of CA rules

Apologies if I come on too strong about this, but it's important to me that we know how things work. I wouldn't want to see you defend this in, for example a job interview, when it breaks the fundamental concepts of the architectural approach that's in question.

There also might be a misunderstanding between us, if you just called this "clean" architecture because of separation of concerns and you didn't have any knowledge of the Clean Architecture approach, then I'm kinda wrong for even mentioning it. Even in that case I think it's good we had this conversation since now you know that naming something "clean architecture" in .NET implies the strict domain-centric structure and should take that into consideration when describing your project. 

Bottom line is, there is nothing wrong with N-Layer structure as it has been used for decades and is widely adopted in .NET companies. I think it's still a very good step forward in your learning process.

## Mapping classes

I've seen both approaches - having a mapping class VS mapping manually inside methods. 

This is really up to you, mapping class introduces some abstraction and not having them gives us code that's a bit more explicit. 

There are multiple approaches on how to achieve this, so I'll leave you with [this link](https://x.com/Dave_DotNet/status/1928792464699121748)

Gincaza: I’ve been thinking about an alternative approach and wanted to check if you think it would be more aligned with Clean Architecture principles.

In this version, inside the BusinessLayer namespace, I would define an interface called ICodingSession, listing the base functions. The BusinessLogicLayer would then interact with the database through a repository interface like:

```csharp
private readonly ICodingSessionRepository _repository;
```
The idea is that **DatabaseManager** would implement ICodingSession, providing the actual logic for those functions. This way, the dependencies would remain similar to the structure shown in the image I attached with both Presentation and DataAccess depending on BusinessLogic.

```csharp
public class DatabaseManager : ICodingSessionRepository
```

image

Do you think this approach would be more acceptable? I'm still learning and trying to understand the best architectural practices.

nwdorian: Yeah, that is exactly how it's structured in CA. 

I strongly urge you to go over video resources I've shared, they specifically adress problems like that and go over much more then we can ever cover in text. All 3 videos combined are 3h long which might seem like much, but you learn this once and you know it forever. It's a worthwhile investment.

nwdorian (approved):
Hey @Gincaza👋,

I have taken a look at the updated application and I will approve the project submission.

Keep up the good work on the next ones! ☺️

PR #217, opened 2025-07-16
Full project implementation for Coding Tracker CLI app

Niasua:


CodingTracker.Niasua/Data/CodingController.cs
```diff
@@ -0,0 +1,239 @@
+﻿using CodingTracker.Niasua.Configuration;
+using CodingTracker.Niasua.Models;
+using CodingTracker.Niasua.UserInput;
+using Dapper;
+using Microsoft.Data.Sqlite;
+using System;
+using System.ComponentModel.DataAnnotations;
```
nwdorian, line 7: 🟠**Unnecessary using directives**

➡️ Remove anything that's not being used. It makes your code less readable and maintainable. 

💡In Visual Studio the compiler will tell you when code isn't being used by slightly fading it. Here's a short video on how to make Visual Studio [automatically remove unused using statements](https://www.youtube.com/watch?v=CC5nohNp7F8)

image

CodingTracker.Niasua/Program.cs
```diff
@@ -0,0 +1,316 @@
...
+            "Delete session",
+            "Delete all sessions",
+            "Update session",
+            "Start stopwatch session",
+            "Filter coding records per period",
+            "Show summary report",
+            "Set coding goal",
+            "Show goal progress",
+            "Exit"
+        }));
```
nwdorian, line 37: 🟠 **Magic Strings**

💡 You should use an Enum to avoid magic strings. Look into the `[DisplayName]` attribute and use a converter with Spectre.Console to show a human friendly version. e.g. `Add Coding Session`.

CodingTracker.Niasua/Program.cs
```diff
@@ -0,0 +1,316 @@
...
+
+            break;
+
+        case "Add new session":
+
+            var session = UserInputHandler.GetCodingSessionFromUser();
+            AddSession(session);
+
+            break;
+
```
nwdorian, line 54: 🟡 **Whitespace**

💡Bit of nitpicking here, I just feel like having a new line surrounding the code in the `case` branch plus another one after `break;` is too much. I do see where you're coming from, it does makes sense, but it also makes the switch-case stretch so much vertically it's a bit hard to read.

README.md
nwdorian, line 1: 🟢 **README File**

⭐ Excellent README file

CodingTracker.Niasua/CodingTracker.Niasua.csproj
nwdorian, line 1: 🟡 **NuGet packages**

💡Lots of people make this mistake, the package is `Microsoft.Data.Sqlite` without the `Core` at the end. Folks think `Core` suffix refers to .NET Core, but it's actually just the bare bones or "the core" of Sqlite package and that's why you had to get `SQLitePCLRaw.bundle_e_sqlite3` which shouldn't be the case.

CodingTracker.Niasua/Data/CodingController.cs
```diff
@@ -0,0 +1,239 @@
...
+
+
+namespace CodingTracker.Niasua.Data;
+
+internal static class CodingController
+{
```
nwdorian, line 13: 🟠 **Static classes**

❓If we make everything `static`, we are throwing away all the OOP concepts, as we no longer work with class instances aka objects.

➡️ As a rule of thumb I would avoid the `static` keyword completely and only use it if you really need it.

nwdorian (approved):
Hey @Niasua 👋,

Excellent work on your **Coding Tracker** project submission! 🎉

I have performed a code review and left some comments on things that I like and also on areas to improve. ☺️

---

🟢 **Requirements**

---

🏅 **Extra challenges**

✅ Stopwatch feature
✅ Filtering coding sessions
✅ Create total and average session reports
✅ Set coding goals

---

🟠 **Solution Structure**

❓ As per the academy's [Code Reviews](https://thecsharpacademy.com/article/52/code-reviews) page, I was expecting your project submission to be contained in a folder with the format: **ProjectName.GitHubUserName**

💡 Just for clarification, you did name your project like requested, but the application is not in a self-contained folder - there should be no extra files or folders in the root of the forked repository except the `CodingTracker.Niasua` folder where you would then add all your projects files and folders.

➡️ Currently in the root of the fork there is a README file, a solution file and your projects folder. What we want to see is just one folder with all of those inside it. This also gives you the freedom to name your projects in a different way, for example `CodingTracker.Console`, with the main root folder still following the academy naming convention, of course.

---

🟡 **Project structure**

❓Overall the structure is more then fine. My small advice is to always strive to keep your `Program.cs` as clean as possible. 

💡I would move the selection menu into a `Menu.cs` class and all other methods into `CodingView.cs` and rename the current controller class (the one that's handling database interactions) into something like `DatabaseMethods.cs` (for now, in later projects we learn more about code structuring patterns). 

➡️ Then in `Program.cs` you would only call db initialization, seeding and run the menu. 3 lines of code - clean, easy and simple.

*ps. if you want to talk about naming in more detail, reach out to me on discord* ☺️

---

🟡 **UX improvements**

➡️ A nice feature to have is the ability to enter "0" at any operation to cancel it completely. For example, when adding a new session, we have no way to exit that operation flow in case we change our mind.
➡️ The menu seems a bit bloated, maybe making a sub-menu for sessions CRUD would be a good idea. Then in main menu have a "Manage sessions" with read, create, delete and update inside it.

---

💪 Wow, really good work on this one, with all extra features you've added this doesn't seem far off from being a real production application. ☺️

⭐ Great job on following the coding conventions aswell, code looks really clean. 

Best regards,  
@nwdorian

PR #221, opened 2025-08-05
ready for review

mosnyan:
have fun!


mosnyan:
codacy is flagging this using directive

using System.Configuration;

removing it obviously prevents the code from compiling due to the configuration file handling. i added the package via nuget, maybe it's the way i did it that it doesn't like?

andymartinez1 (changes requested):
Hi @mosnyan. Excellent job on this project and the project structure! Don't worry about the Codacy issue. That seems to happen with configuration imports.

### Feedback:
- The duration doesn't seem to be calculating correctly. Only the hours are being calculated. Here are some examples:
Duration-Time

Please fix this so it can be approved as per this requirement:
image

### Smaller Issues:
- If I delete coding session Id 1, coding session Id 2 then becomes Id 1. 

- Another thing, though small, it would be nice to have some feedback when selecting "See all sessions" when the list of coding sessions is empty.  

Please reach out if you have any questions!

mosnyan:
oops, silly me, i fixed the formatting issue so days should show up now

andymartinez1 (approved):
Excellent job! I think you're very well prepared for the intermediate projects.

PR #227, opened 2025-08-28
Coding Tracker project completion

JJHH17:
The project:
- Allows user to add coding time manually.
- Allows user to add coding time via a stopwatch/live tracking.
- View all entries.
- Delete a single entry.
- Delete all entries.
- Pulls database strings via app.config file.


VocalNight (commented):
Good job on the project.

Parsing Datetime will get easier, but it's definetly something you have to get used to.
I do recommend learning Regex either way, sicne you mentioned it. Check out this site https://regexone.com/

I will be approving the project, good job!

PR #229, opened 2025-09-24
Finished project

KacperZys:
I tried to use MVC architecture. It's been a fun project.

### DIFF
--- a/CodingTracker.KacperZys/CodingTracker.KacperZys.sln
+++ b/CodingTracker.KacperZys/CodingTracker.KacperZys.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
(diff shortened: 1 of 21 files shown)


TheCSharpAcademy:
@KacperZys Project approved! 😄✅ The UI looks really good. And well done for completing the challenges! Each little extra task teaches you something different that you might use later in your career. Keep up this good habit!

🥋And congratulations! You got your OLIVE GREEN belt!! 🟩🟩

🔍️To be honest I don't have a lot of feedback, your code is clean and tidy and you follow good principles (SoC, DRY, etc). You seem to have understood the basic principles of OOP. Keep up the good work.  

🐞One thing I suggest for the future is to always handle errors somehow when dealing with databases. You could wrap the code in your CRUD methods with a try-catch block so the app doesn't crash if there's an exception. This will be a requirement in later projects, but I suggest you start thinking about it straight away. 💪💪

PR #231, opened 2025-10-06
Project finished!

RSilva9:


TheCSharpAcademy (changes requested):
@RSilva9 Thanks for submitting! The project is looking great but I found one issue. You shouldn't allow end date to be before start date:

image

RSilva9:
Hey! Thanks for taking the time to review my project!

Do you mean End Time? I figured the user could start coding at a certain hour and finish their session after midnight, making a day pass. So it wouldn't be an earlier hour but an earlier time of the following day.

Hope I was able to explain properly.

TheCSharpAcademy:
@RSilva9 Makes sense. In a real app you'd have to be more explicit about this, either explaining this to the user or adding the end date along with the time when viewing records. 

So for now... Well done! Project completed! ✅The code is super clean and organized. And well done for completing the challenges! 🤓 Good work! 👏👏

Oh just one more thing. I can't find the "sarddo" username on Discord. Are you sure it's correct? 

See you in the Flashcards project 🤓

PR #239, opened 2025-12-14
My attempt at the CodingTracker project!

diegopetrola:
Hello!

This is my attempt at the CodingTracker project. I believe it achieves everything that is asked, including challenges, with the exception of the part of setting the goal.

I tried to use the Controller, Service, Data pattern that is ubiquitous among dotnet projects, specially web ones, but I don't know if it's being used perfectly.


CodingTracker/Controller/CodingSessionController.cs
```diff
@@ -0,0 +1,247 @@
+﻿
+using CodingTracker.Data;
```
nwdorian, line 2: 🟡 **Whitespace**

❓There should be no empty lines at the top of the file. 

➡️ Set up Format on save in your IDE to reduce the chance of forgetting about formatting.

CodingTracker/Controller/CodingSessionController.cs
```diff
@@ -0,0 +1,247 @@
...
+
+public class CodingSessionController : ICodingSessionController
+{
+    private readonly ICodingSessionService service;
+    private readonly JokeService jService = new ();
+
+    public CodingSessionController(ICodingSessionService service)
+    {
+        this.service = service;
+    }
```
nwdorian, line 18: 🟠 **JokeService interface**

❓ If you already used interfaces for other layers, you could've extracted an interface for the JokeService and injected it like others.

➡️ You are instantiating everything in Program.cs except this one service in here. Keep it consistent 😄

🟠 **JokeService field name**

❓ Following [.NET naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#using-abbreviations-and-acronyms) we shouldn't use abbreviations or contractions as part of identifier names.

➡️ Just name it `jokeService` - you didn't really achieve much by not having a full word there.

CodingTracker/Controller/CodingSessionController.cs
```diff
@@ -0,0 +1,247 @@
...
+        }
+        catch (Exception e)
+        {
+            AnsiConsole.MarkupLine($"\n\n[red]{e.Message}[/]\n\n");
+            return;
+        }
+        AnsiConsole.MarkupLine("\n\n[green]Session inserted successfully![/]\n\n");
+    }
+    public async Task DeleteSession()
+    {
```
nwdorian, line 49: 🟡 **Whitespace**

❓ Always add a single line of whitespace between the end and start of methods. 

➡️ As mentioned before, some automated formatting feature should take care of this for you, so you don't forget. You have this problem in other classes, I won't point it out everywhere. Look into CSharpier in combination with an `.editorconfig` file for rules. Reach out on discord if you need help setting this up.

CodingTracker/Controller/CodingSessionController.cs
```diff
@@ -0,0 +1,247 @@
...
+            .Width(60)
+            .Label("[green bold underline]CODING REPORT (in hours)[/]\n\n")
+            .CenterLabel();
+
+        barChart.UseValueFormatter((val) => val.ToString("N"));
+
+        for (int i = 0; i < reports.Count; i++)
+        {
+            barChart.AddItem(reports[i].Period, reports[i].Duration.TotalHours, colors[i % colors.Length]);
+        }
```
nwdorian, line 175: 🟡 **Use Foreach**

❓ When you are just looping from start to end of the collection use a `foreach` loop, it's simpler and more readable.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+            .AddJsonFile("appsettings.json").Build();
+
+        ConnectionString = _config["ConnectionString"]
+            ?? throw new Exception("Please configure a valid connection string");
+    }
+
+    public IDbConnection CreateConnection()
+    {
+        // For a local database I supposed there is no harm in creating a new
+        // connecton for each query, it might even be good to avoid locking the db file.
```
nwdorian, line 25: 🟡 **Code comments**

❓ It's fine to add comments during development if it's going to help you with pseudo-code or to remember some specific logic. When you finish a project, always go back through the code for a polish pass and clean up unnecessary comments.  

💭 Remember, our code should be self-documenting - with good naming and structure we shouldn't need any comments at all. If you really want to point something like this out (assuming it's for reviewers), make notes as you are working and add it to PR description.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+        db.Execute(createTable);
+
+        if (seedData)
+            await SeedData();
+    }
+
+    public async Task SeedData()
+    {
+        using IDbConnection db = new SqliteConnection(ConnectionString);
+        int count = db.QuerySingle<int>("SELECT COUNT (*) FROM CodingSession");
```
nwdorian, line 49: 🟡 **Use `CreateConnection()`**

❓ Reuse the method you created earlier. I guess this is something to think about - manually initializing the SqliteConnection versus wrapping that one line in a method doesn't give you much benefits.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+using Microsoft.Data.Sqlite;
+using Microsoft.Extensions.Configuration;
+using System.Data;
+
+namespace CodingTracker.Data;
+
+public class CodingSessionData : ICodingSessionData
+{
+    private readonly IConfiguration _config;
+    private string ConnectionString { get; set; }
```
nwdorian, line 11: 🟠 **Private fields naming**

👉 Following [.NET naming conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions) private instance fields should start with an underscore (`_`) and the remaining text is camelCased. Just like you have the `_config`

👉 I'd also argue that this shouldn't be a property. Property implies it will change, but you set the connection string once and it never changes - it should just be a `private readonly string _connectionString;` field.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+        db.Execute(insertSql, seedData);
+    }
+
+    public async Task InsertSession(CodingSession session)
+    {
+        using var conn = CreateConnection();
+        var sql = "INSERT INTO CodingSession (StartTime, EndTime) VALUES (@StartTime, @EndTime)";
+
+        await conn.ExecuteAsync(sql, session);
+    }
```
nwdorian, line 87: 🟠 **Exception handling**

❓ Looking at other layers, it seems you weren't really sure where to place try-catch blocks for exception handling. Simple guideline is to place them as low as possible. 

➡️ Code that executes or queries with SQL should have try-catch blocks. Just let them bubble up to upper layers. This is the case for all methods in this file. This also clears up upper layers and leaves focus on logic and display instead of polluting the code in services and controllers with try-catch blocks.

💭 I guess a valid reason for not placing it here is to not have data layer deal with displaying messages, but unless you use some form of Result pattern you can't really avoid this.

```suggestion
    public async Task InsertSession(CodingSession session)
    {
        try
        {
            using var conn = CreateConnection();
            var sql =
                "INSERT INTO CodingSession (StartTime, EndTime) VALUES (@StartTime, @EndTime)";

            await conn.ExecuteAsync(sql, session);
        }
        catch (Exception ex)
        {
            AnsiConsole.MarkupLine($"[red]{ex.Message}[/]");
        }
    }
```

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+    public required DateTime StartTime { get; set; }
+    public required DateTime EndTime { get; set; }
+    public TimeSpan Duration => EndTime - StartTime;
+}
+public class CodingReport
+{
+    public string Period { get; set; } = "";
+    public long TotalSeconds { get; set; }
+    public TimeSpan Duration => TimeSpan.FromSeconds(TotalSeconds);
+}
```
nwdorian, line 174: 🟠 **Model placement**

❓ Place models into Models folder and create a file per model. This is a widely accepted convention in .NET

CodingTracker/Services/CodingSessionService.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace CodingTracker.Services;
+
+public class CodingSessionService : ICodingSessionService
+{
+    ICodingSessionData data;
+
+    public CodingSessionService(ICodingSessionData data)
+    {
+        this.data = data;
+    }
```
nwdorian, line 14: 🟠 **Private fields naming**

👉 Following [.NET naming conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions) private instance fields should start with an underscore (`_`) and the remaining text is camelCased.

➡️ Using the `this` keyword is outdated and having different names for the constructor parameter and the private field enables you to omit it while keeping the intention explicit.

➡️ Also when you have multiple classes for different data models you will not be able to just use `_data` - keep the full name of the class.

```suggestion
    private readonly ICodingSessionData _codingSessionData;

    public CodingSessionService(ICodingSessionData codingSessionData)
    {
        _codingSessionData = codingSessionData;
    }
```

CodingTracker/Services/CodingSessionService.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+        catch (Exception e)
+        {
+            throw new Exception("Database error: ", e);
+        }
+    }
+    public async Task<List<CodingSession>> GetAllSessions()
+    {
+        return await data.GetAllSessions();
+    }
```
nwdorian, line 30: 🟠 **Exception handling**

👉 Every method that touches the database needs to be wrapped in a try-catch block. It seems you only wrapped the methods related to creating and deleting, but what happens if db connection breaks during reading ? You get an uncaught exception and your app breaks. 

➡️ As mentioned earlier just wrap all db interaction code in try-catch in the data layer.

CodingTracker/Services/JokeService.cs
```diff
@@ -0,0 +1,126 @@
+﻿namespace CodingTracker.Services;
+
+public class JokeService
+{
+    static List<Joke> jokes = [];
+    private Random rand = new();
+
```
nwdorian, line 7: 🟠 **Access modifiers**

❓ Always be explicit with access modifiers, don't omit them. Even thought it's private by default you should still clarify intent.

CodingTracker/Services/JokeService.cs
```diff
@@ -0,0 +1,126 @@
...
+        jokes.Add(new Joke
+        {
+            Question = "Why shouldn't you trust Matlab developers?",
+            Answer = "Because they're always plotting something."
+        });
+    }
+    public Joke GetJoke()
+    {
+        return jokes.ElementAt(rand.Next(jokes.Count));
+    }
```
nwdorian, line 117: 🟠 **Random instance**

❓Use `Random.Shared` static instance for most scenarios, especially simple random selections like picking ints. Use an instance field when you need deterministic behaviour for testing or when working in performance-critical, high-concurrency situations.
```suggestion
    public Joke GetJoke()
    {
        return jokes.ElementAt(Random.Shared.Next(jokes.Count));
    }
```

CodingTracker/appsettings.json
nwdorian, line 1: 🟢 **Using appsettings.json**

✅ Good job on using appsettings.json, it's best practice that is utilized in modern .NET

CodingTracker/Services/JokeService.cs
```diff
@@ -0,0 +1,126 @@
...
+
+}
+
+
+public class Joke
+{
+    public string Question { get; set; } = "";
+    public string Answer { get; set; } = "";
+}
```
nwdorian, line 126: 🟠 **Model placement**

❓ Same as previous models, this should be in a `Joke.cs` file in `Models` folder.

CodingTracker/CodingTracker.db
nwdorian, line 1: 🟠 **Database in Repo**

💡 You shouldn't be committing your database file to source control. Add it to the `.gitignore`.

CodingTracker/InputValidator.cs
```diff
@@ -0,0 +1,87 @@
...
+                        Today - [bold]units[/] where [bold]units[/] is [bold]d[/]ay, [bold]w[/]eek and [bold]m[/]onth respectively.
+                    3. Type [bold]now[/] to get the current time
+
+                Example:
+                    -5d => Today - 5 days
+                    10w => Today + 10 weeks
+                    5 m => [red]invalid input[/], the number and the character must be together!
+
+                """);
+    }
```
nwdorian, line 28: 🟠 **Code placement**

❓This doesn't do any validation, it's better to place it in `UserInterface`. The `dateFormat` property is public so you can still interpolate it in another class.

CodingTracker/InputValidator.cs
```diff
@@ -0,0 +1,87 @@
...
+            return dateString[^1] switch
+            {
+                'm' => DateTime.Now.AddMonths(val),
+                'w' => DateTime.Now.AddDays(val * 7),
+                _ => DateTime.Now.AddDays(val),
+            };
+        }
+
+        return DateTime.ParseExact(dateString, dateFormat, CultureInfo.InvariantCulture);
+    }
```
nwdorian, line 74: 🟠 **Validation**

1️⃣ They say if you solve a validation problem with regex you have two problems: validation and regex. There's always a way to not use regex so try and find ways around it. 

2️⃣ Overall this validation code is quite complicated and hard to follow. It would help a lot if you extract validation into it's own class - so break this logic up into 2 classes like `InputValidation` and `UserInput` - you will also thank yourself later when you have to write unit tests for this project 🙂 

3️⃣ Spectre.Console prompts are generic so you can have `new TextPrompt<DateTime>` that does tryparse and returns the DateTime type you really want. Of course you have the letter shortcuts but that could also be solved with separate methods.

4️⃣ Maybe have a look at the [examples ](https://discord.com/channels/925901888203935744/1143303759804498012/1311041643704615014) I wrote in community discord. The key is that `.Validate()` takes a `Func` so you can just put methods with correct signatures in there.

CodingTracker/InputValidator.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    Week,
+    Month
+}
+
+public enum ReportGrouping
+{
+    Total,
+    Average,
+}
```
nwdorian, line 87: 🟠 **Enum placement**

❓ Same as with models, unless these are private - put them in an `Enums` folder with an enum per file structure.

CodingTracker/Program.cs
nwdorian, line 1: 🟢 `Poor man's DI`

💪 Seeing this just makes my heart warm ❤️ This is absolutely the best way to start with DI, one project with poor man's DI and then use a proper service collection in next ones. 

✅ I'm also happy to see you implemented lessons from the DI course, good job!

CodingTracker/UserInterface.cs
```diff
@@ -0,0 +1,33 @@
...
+        Dictionary<string, Func<Task>> menuActions = new()
+        {
+            { "Start Live Session", controller.LiveSession },
+            { "Add Session", controller.AddSession },
+            { "Delete Session", controller.DeleteSession },
+            { "Update Session", controller.UpdateSession },
+            { "Filter Sessions", controller.FilterSessions },
+            { "Coding Report", controller.CodingReport },
+            { "Hear a Joke", controller.HearAJoke},
+            { "Quit", () => { Environment.Exit(0); return Task.CompletedTask; } }
```
nwdorian, line 19: 🟠 **Magic Strings**

➡️ You should use an Enum to avoid [magic strings](https://deviq.com/antipatterns/magic-strings). You can create a simple helper method to show a human friendly version of the Enum in the menu.

💡Again, a reminder that Spectre text prompt can be of other types then strings. I think checking out an [example ](https://github.com/nwdorian/ExerciseTracker/blob/develop/src/ExerciseTracker.Console/Features/Categories/CategoriesMenu.cs) from my project can give you some inspiration. It uses a switch statement instead of a dictionary, but I can also make a case that this is not _the best_ use of a dictionary - just because you use a dictionary to have multiple values per key and you are basically doing string to method mapping which is perfect for `switch`.

CodingTracker/Data/CodingSessionData.cs
nwdorian, line 1: 🟠 **Single responsibility principle**

➡️ Othen than extracting models into files you could also break the class up into at least two files. Classes that handle database interaction with a model are commonly called repositories. My suggestion:

**Data** - folder
- `DatabaseInitializer` - `CreateDatabase` and `SeedData` methods
- `CodingSessionRepository` - all methods related to CodingSession model
- `CodingReportRepository` - all methods related to CodingReport model (even if theres only one)

This also reduces the number of lines per file which helps with readability and maintainability of your code.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    public async Task UpdateSession(CodingSession session)
+    {
+        using var conn = CreateConnection();
+        var sql = @"
+            UPDATE CodingSession
+            SET StartDate=StartDate, EndDate=@EndDate
+            WHERE ID=@ID";
+        await conn.ExecuteAsync(sql, session);
+    }
```
nwdorian, line 115: 🟠 **Update doesn't work**

🐞 Your table columns are `StartTime` and `EndTime` - you were probably in a hurry and misspelled these.

➡️ Make sure to thoroughly test the application before submitting for review. Cover happy and failure paths, invalid inputs and edge cases. Code reviews get stricter in future projects and you might be asked to fix issues like this.

CodingTracker/Data/CodingSessionData.cs
```diff
@@ -0,0 +1,174 @@
...
+            Console.WriteLine($"{report.Period} - {report.Duration.TotalHours:N1}");
+        }
+
+        return reportData.ToList();
+    }
+}
+
+public class CodingSession
+{
+    public int ID { get; set; }
```
nwdorian, line 164: 🟠 **Naming**

❓ Following [.NET naming conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalizing-compound-words-and-common-terms) we should treat acronyms and compound terms as words.

👉 The property should be named `Id` instead of `ID`

nwdorian (approved):
Hey @diegopetrola 👋,

Excellent work on your **Coding Tracker** project submission! 🎉

I have finished the code review and left some feedback below. ☺️

Feel free to reach out if you have questions or if you'd like to collaborate on any of these points.✌️

---

🟢 **Requirements**

⭐ You have completed all of the project requirements!

---

🏅 **Extra challenges**

✅ Stopwatch feature
✅ Filtering coding sessions
✅ Create total and average session reports

---

🟠 **Submission Structure**

❓ As per the academy's [Code Reviews](https://thecsharpacademy.com/article/52/code-reviews) page, I was expecting your project submission to be contained in a folder with the format: **ProjectName.GitHubUserName**

👉 For example, `{ProjectName}.{GithubName}` with all application related files and folders inside it. The submission should be in self-contained top-level folder with correct naming. Take a look at this structure for example:

📁 `CodeReviews.Console.CodingTracker` ⬅️ Forked repo
 ┣ 📄 `.codacy.yml`
 ┣ 📄 `.gitignore`
 ┣ 📁 `CodingTracker.diegopetrola` ⬅️ Top-level folder
 ┃  ┗ 📄 `CodingTracker.sln`
 ┃  ┗ 📄 `README.md`
 ┃  ┗ 📁 `CodingTracker`

💡The easiest way to tackle this problem is to create a new folder for your submission right after cloning the fork. This will make it easier for code-reviewers to review your code in the future. It will also make sense when you get to the projects which contain other students code.

---

🟡 **UX improvements**

➡️ Whenever you are designing console applications, make sure to clear the console after each operation or it gets cluttered pretty quickly. Add `Console.Clear()` at start or end of the while loop.

➡️ When doing Delete and Update operations, display a list of records with visible Ids. Currently you have to remember the record Id you want to edit/delete. 

➡️ It's not necessary to ask for seconds when entering date and time. It makes the user have to do more for little benefit.

➡️ You have `dateFormat` in `InputValidator` - use that to display uniform formats across the app. I'm required to input date in one format but it's displayed to me in another (this probably wasn't the case on your PC and it's likely caused by different locales used by our machines)

---

✅ Good job taking care of error handling and coding conventions, file-scoped namespaces and using declarations are amazing features that are often overlooked. 

💭 I left a lot of comments but it's all with good intention. Please don't get discouraged by any of the feedback - I was splitting hairs on some minor things. ☺️

⭐ You did an amazing job implementing the extra challenges which is not always simple. 

---

💭 You can take care of issues in your own repo if you want, but since the project will be approved you don't have to push the changes or create a new PR.

Best regards,  
@nwdorian

diegopetrola:
Thank you very much for your feedback, nwdorian. It touches on many questions I had while designing the system for this project. I will try to incorporate your insights on the next one (that is already in the works).

I will also solve the update bug here because I want to make sure everything is working properly at least.

PR #240, opened 2025-12-15
Finished the coding tracker

TerrenceLGee:
This was kind of tough because I wanted to add more and more functionality and I'm not sure if I implemented this in the best way, but I did give it my all. Also incorporated Unit Tests.


TheCSharpAcademy:
@TerrenceLGee  Project approved! 😄✅ I have no words, your project is just on a different level. You left no stone unturned. I'm particularly impressed with your unit tests, you covered every base. Please keep these standards, it truly stands out. You should definitely create a short video of this app and [display it in our gallery](https://thecsharpacademy.com/Gallery).

🥋And congratulations! You got your **Olive Green Belt**!! 🫒🫒 

🔍I tried hard to find something to give you feedback and the only thing I could find was that you could add more indentation and line breaks to make some of your SQL queries more readable. But I understand it's a matter of personal preference. 

image

image

TerrenceLGee:
Wow!!! Thank you so very much for your kind words and praise!!! I honestly thought I did too much and overdid things 😄. It took me roughly two weeks to do this and I actually did it about 3 times before I was satisfied enough to submit it, and even then I was unsure if it was good enough😄.   I hope that I can keep improving and I know that I will following the road-map! Thank you again for your kind words and praise it means so much to me and thank you for the wonderful resource that is The C# Academy!

PR #241, opened 2025-12-23
Review

filipmarkovic064:
Tried my best on this one, be extra strict please! 
Thanks in advance :)


filipmarkovic064:
Uh oh Codacy is detecting 19 issues and all of them are the automatically generated "Using ..." when using libraries, i tried removing them but the program doesnt work when i remove them so ill just leave it be, I checked on discord and people are saying that it will get reviewed anyway

andymartinez1 (changes requested):
HI @filipmarkovic064! Great job on this project! Don't worry too much about the Codacy thing as it tends to detect issues that cannot be fixed without breaking the code. The UI is clean and easy to navigate. However, it is missing a requirement:

From the Coding Tracker project
image
From the Habit Logger project
image

There needs to be a way to update user submitted coding session entries in case the user enters incorrect dates.

Since you asked for a strict review, I found two things that can be improved though they are not requirements.
1. It would be nice to be able to go back from this screen when selecting a coding session to delete rather than having to select one and selecting no for the confirmation (The confirmation was a very nice addition by the way)
image

2. I recommend looking into converting enums into strings using display name for display purposes. So instead of the menu showing a long single word of "SortedByDurationDescending", it'll show "Sorted By Duration Descending" for example.

Overall, you did a very good job on this project so far. Especially with completing the challenges as well. Please add an update feature so I can approve the project. I look forward to seeing the changes!

filipmarkovic064:
Hey thanks for the thorough review! 
Added the update functionality (i made it so users have to input both the start and end time again because i was a bit lazy :D)

Made it so you can cancel deleting or updating a session. This feature was never implemented into Spectre Console (https://github.com/spectreconsole/spectre.console/discussions/700) so i had to make my own workaround.

Adding description to Enums was surprisingly difficult but i got it done using some dictionaries and .UseConverter from SpectreConsole

andymartinez1 (approved):
Excellent job with the changes! This project came out great. I will approve and leave a few considerations for future projects:
1. Organize files into directories (e.g. UserInterface.cs into a UI or Views folder, CodingSessions into a Models folder, etc.)
2. Consider using an Enum extension class for converting enums to strings. It'll be used in almost all future projects.
3. If you're using Visual Studio, there's a code cleanup button. If you're using Rider, you can reformat on save. Just to keep code format consistent throughout the entire project.

You're ready for the intermediate projects now. It seems like you have the logic down so it should be fairly simple. I look forward to seeing your next projects!

PR #243, opened 2025-12-28
My Coding Tracker Console Application

schoffen:
Hello there,

I’ve just finished this project, and it was even more challenging and rewarding than the previous one.

I felt quite lost at times while working with separation of concerns. I did my best to structure everything correctly, but if something is not right, I’d really appreciate your feedback. This was, by far, the hardest part of the project for me, and also the one where I learned the most.

Thank you in advance for taking the time to review it.

Kind regards,
Felipe Schoffen

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,29 @@
+# Coding Habit Tracker
+This is a coding session tracker console application. It was made with C#, Dapper and Spectre.Console.
+
+### Requirements:
+- This application must log daily coding time, using specific format;
+- This application must use Spectre.Console library;
+- Use separation of concerns;
+- Use configuration file for database path and connection strings;
+- The user must only input start date time and end date time, the application is responsible for calculating the duration;
+- Use Dapper ORM for the data access;
+- Follow DRY Principle;
+- This READ ME file.
+
+### Optional Requirements:
+- Possibility of tracking time via stopwatch;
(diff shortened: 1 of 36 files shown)


TheCSharpAcademy:
@schoffen Project approved! 😄✅ Well done for completing the challenges! 🥋And congratulations! You got your Olive green belt!! 🫒🫒 

👩‍🏫I'm glad you found the Separation of Concerns requirement useful. This is where things start to get a little bit more complicated because it's not just about getting your code to do the things that you want it to do, but also about how to make your code more organized. That's a key aspect of a developer's skill set, particularly when working in teams.

**Feedback**
🔍️Your code is clean and tidy and you follow good principles (SoC, DRY, etc). You showed good command of OOP principles and the app does everything it's supposed to do. Excellent work 😀.

🔍️For the next project I highly recommend you start using unit tests. You don't need to cover your entire code, but this is only going to get more important as your projects grow in complexity and it will save you a lot of headaches. 😀 [Here's a good place to get started](https://[thecsharpacademy.com/article/30009/unit-testing-for-beginners).

🔍️Something else you need to start thinking about is: What happens when you have 1000 records? Will you retrieve them all in a wall of text? 😀 This is what we call **"pagination"** where you retrieve only a subset of records (a page). GetAll is only a good idea in small not user-defined tables. 

image

✋🏻Overall fantastic work. Keep up the good standards! Looking forward to seeing your next projects!

PR #246, opened 2026-01-02
Coding Tracker Implementation

aneevel:
See README.md for a detailed description!


TheCSharpAcademy:
@aneevel  Project approved! 😄✅ And well done for again completing the challenges! I'm glad you're including unit tests. I suggest you always do that during development from now on. 👌 Also your Readme is excellent! Much better than just a list of technologies/instructions. It helps folks to have a feel for how you think and helps you develop your communication skills. Highly recommend you keep that habit too 😎

🥋And congratulations! You got your Olive Green belt!! 🟩🟩

**Feedback**
🔍️The only thing I'd absolutely have done differently is the number of layers. Your service layer is pretty empty at the moment, adding unnecessary complexity. 

🔍️I like the idea of handling errors the way you did, adding some sort of stack trace for better logging/debugging. However I think I would have added a **global exception handler** at the app's entry point and let it bubble up. That's just because the exception handling is taking 80% of the space in your repository and I think it could have been handled in a less verbose way.

🔍️Some aspects of the app do feel a bit convoluted as you stated in your Readme (even though I didn't dig too deep). The Flashcards app coming next can get complex really quickly due to table relationships so I recommend you try a KISS (keep it simple stupid) approach. 

✋🏻Overall great job! Looking forward to seeing your next projects!

PR #247, opened 2026-01-06
CodingTracker

NoNameLudovik:


andymartinez1 (approved):
Hi @NoNameLudovik. Excellent job on this project! Everything works as expected and the README was detailed and helpful. 

Feedback:
I found 2 very minor things that could use some improvement.
1. I'd recommend adding a prompt such as "Press any key to continue.." after the menu selections so the user knows how to move to the next screen
image

2. Consider converting the enums to strings using a display name enum extension so the menu options read as "Add Session" rather than "AddSession" for instance.

Great job overall, and I look forward to seeing your next projects!

PR #248, opened 2026-01-08
Complete CodingTracker

coolorvi:


andymartinez1 (changes requested):
Hi @coolorvi. Great job on this project! The project structure was well done and you handled separation of concerns well. Most of the requirements were met, however, I found two issues that need to be resolved. Please fix them so I can approve the project.

1. The README should also include your thought process behind this project. Please add some more detail as well and don't forget to include how to set up the project and run it. The user needs to modify the launchSettings.json to run the app. That needs to be included as well. 

2. When entering the same time on a different day, it shows 0 hours for duration:
image


<hr/>
There are also some other things I think could use some improvement though they are not requirements for me to approve the project:

- Consider validating the user input immediately after the first input instead of after entering both the start and end time
{5DA1F31B-E0ED-4C31-8607-79BBF9D28C55}

- Another consideration: The console gets cluttered after a while. When going back to the main menu, it might be a good idea to use an AnsiConsole.Clear(). Ex:
```csharp
case "Create a new session":
    cc.CreateSession();
    AnsiConsole.Clear();
    break;
```

coolorvi:
Thanks for review. I fixed the error of calculating a session lasting more than one day, rewrote the README and corrected launchSettings.json.

andymartinez1 (approved):
Hi @coolorvi, great job with the changes! Everything is working as intended now. Of course, coding for 24 hours is unlikely but it's good to figure out how to fix that issue now before you need to use it for a future project. 

I look forward to seeing your next projects!

PR #249, opened 2026-01-10
Code Tracker Project

davetn657:
Hi thanks for taking the time to review my code!


andymartinez1 (approved):
HI @davetn657. Great job with your project! Everything seems to be working correctly, you included a good README and the project structure is well done.

I will approve the project but I did find a few things that could use improvement. Please consider these things for future projects or if you go back to this one later on.
1. Adding seconds is just more typing the user has to do. Try to simplify it by using only hours and minutes
image

2. Adding a confirmation such as "Coding Session Created" or "Coding Session Deleted" would be nice instead of having to go back to See Existing Records to see the changes.

3. Adding a way to cancel adding/updating or getting a report would also be nice. The way you have it now, you have to complete the add or edit:
image

4. The filtering options aren't working. No matter what date format I use, it asks for a new one

5. There is validation but you should include something along the lines of "Wrong date format used, please try again" then another prompt. The way you have it now, it just asks for a date input again which can be confusing.

Overall, you did a great job and I look forward to seeing your next projects.

PR #250, opened 2026-01-16
Completed Project!

ktawiah:


TheCSharpAcademy:
@ktawiah Project approved!😄✅  Apologies for the delay in reviewing your submission. We've been flooded with projects recently. Even though you haven't completed the suggested challenges, adding a log table is a very nice "real world like" addition to your project. Very useful! And the UX/UI are super slick with Spectre.Console 👏👏 

**Feedback**
🔍️The main issue with your project is that at the moment deleting a session crashes the app because it's a foreign-key associated with it. You'd need to configure your relationship with a "ON DELETE CASCADE" constraint to prevent that. 
<br>
image
<br>

🔍At the moment the **Service>Repository layers** seem to be an overkill. All the service is doing at the moment is wrapping one-line calls to the repository. It's ok if you wanted to practice this pattern, but in real life I'd only add an extra layer when absolutely necessary.

🔍By convention, .NET/C# devs use four lines of indentation instead of two.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

🧑‍🏫Overall well done! 👏👏 By the way great Readme!!✋🏻Looking forward to seeing your next projects!

PR #252, opened 2026-01-28
Initial submission

ghost:
Thank you :)

### DIFF
--- a/CodingTracker.m1chael888/CodingTracker.m1chael888.slnx
+++ b/CodingTracker.m1chael888/CodingTracker.m1chael888.slnx
@@ -1,0 +1,3 @@
+<Solution>
+  <Project Path="CodingTracker.m1chael888/CodingTracker.m1chael888.csproj" />
+</Solution>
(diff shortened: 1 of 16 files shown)


andymartinez1 (changes requested):
Hi @m1chael888. Your project is looking good so far! I do have one thing that needs to be changed in order for me to approve the project:

Image

That calculation is wrong. Of course nobody codes for 34 hours straight (at least hopefully not) but the code should reflect the correct duration.

<hr>

Some other things to consider as well:
- The user shouldn't have to enter seconds too. That seems like a bit too much
- I would suggest using (YYYY-MM-DD) since this is the international standard for representing dates and times (ISO 8601).
- Consider using an enum extension method to convert your menu options to strings so options look cleaner. 
Example: AddSession would look much better as "Add Session"

<hr>

Please make the change to the duration calculation so I can approve the project. Everything else looked great!

ghost:
Hi, thank you for the feedback.

Ive implemented a fix for the duration calculation and taken your advice with the date formatting/input validation and enum extension method for better menu options. Ive used the same extension method in my flashcards app, its very helpful. I also realized I forgot to handle an empty session list and added in precautions. 

Thanks again =)

andymartinez1 (approved):
Hi @m1chael888. Great job with the updated code! I am approving this project. I look forward to reviewing your future projects!

Best regards,
@andymartinez1

PR #256, opened 2026-02-16
Coding Tracker Project Finished

jcshepherd63:
This project was definitely more difficult that the Habit Tracker project; however, it forced me to figure out how to do research/read documentation which was very beneficial.


andymartinez1 (changes requested):
Hi @jcshepherd63. While reviewing your project, I noticed that there is no duration for the coding session. This is one of the requirements of the project:

Image

Image

<hr>

Everything seems to be working well, but there are a lot of inconsistencies in your code and formatting. 

Image

Missing namespace:

Image

I would suggest going through your code and doing a cleanup and reformat before every project submission. 

<hr>

⚠️ For DateTime format:

I would suggest using (YYYY-MM-DD) since this is the international standard for representing dates and times (ISO 8601).

<hr>

I know some of these may sound like nitpicking, but I am trying to set you up for success with the upcoming projects. They will become harder and reviews will become more strict.

Please add the missing requirement so I can approve.

Best regards,
@andymartinez1

jcshepherd63:
@andymartinez1 Hi Andy! I genuinely appreciate the strict feedback! That is the only way I will improve correctly. I have fixed the date formatting, added the duration column, removed the unused using statement, and added namespaces. Thanks!

andymartinez1 (approved):
Hi @jcshepherd63. Excellent job with the fixes! Everything is working correctly. I am approving the project as all requirements are now met. I do have some more feedback though:

<hr>

I thought this was a very cool addition. A lot of students only do the bare minimum so seeing something like this is very nice.

Image

<hr>

Other Feedback:
⚠️ When updating or deleting, it asks for the input before checking if that record exists. Example:

Image

It says deleted but there was never a session with the ID of 5. Same happens when updating but then it doesn't update anything.

⚠️ You're using Enums which is great but to take it a step further, look into converting enums into strings using a converter in an enum extension method. This will help you go from having menu options like "Add_Coding_Session" to "Add Coding Session" so it looks a little more professional.

<hr>

I look forward to reviewing your future projects!

Best regards,
@andymartinez1

PR #260, opened 2026-03-03
Project Finished!

luc-me:


andymartinez1 (changes requested):
Hi @luc-me. I am trying to review your code but there are some issues with inserting a record:

Image

Also, entering 0 does not return to the main menu:

Image

Deleting a record doesn't work either. 

<hr>

Please review the project requirements and fix the issues so I can finish reviewing it.

Best regards,
@andymartinez1

luc-me:
The delete function it works fine, in my pc at least. 
image
image

Now the function to return to main menu typing 0 it works, I forgot to added 😅.

And for the first issue that you can't insert a record with a later time than the actual time, is a input validation I added but if its wrong iI'll remove it. I was thinking it doesn't make sense add a session of coding before finishing.

Sorry for my english and thanks you for the feedback.

andymartinez1 (approved):
Hi @luc-me. It is working correctly now. I agree it doesn't make sense to add an end time before the start time but users will do stuff that you will question all the time. You should always code as if the app would go into production and be used by many users. Even on a small project like this.

You were right, the delete does work correctly. I misread and didn't select it before deleting. Sorry about that. 

<hr>

Overall, you did a great job! Please consider my feedback as constructive input intended to help improve the project rather than as criticism. I’m approving this project and look forward to reviewing your future work.

Best regards,
@andymartinez1

PR #264, opened 2026-03-24
Finished project including one of the challenges

stevenwardlaw:


TheCSharpAcademy:
@stevenwardlaw  Project approved! 😄✅ And well done for completing the challenges! I really like the multiple options for sorting and filtering. Very well played sir!

🔍️From now on I strongly recommend you create unit tests for your projects. When we initially created the academy this was the latest project of the console area, but with AI tests are much easier to set up, so it will be a great add early on in your journey. In this specific project a great candidate for unit tests would be your validation class. Keep that in mind moving forward!

🔍️When updating records it would be a great if the users could update only one property. It's a bit annoying to have to repeat both dates. You could also add a one-click option for "now" in case the session is finishing now. This will be even more important if the row you want to update has 3+ properties. 

Other than that the code is pretty neat!✋🏻Looking forward to seeing you in the intermediate console area!

PR #265, opened 2026-04-09
Added my project

Ledana:
I followed the tutorial on youtube from csharpacademy then i added the possibilty that the user puts the time they started and ended coding and duration is calculated. i used Dapper to acces the Sql Server database and i also included the code to create the db if it doesnt exist


TheCSharpAcademy:
@Ledana  Project approved.😄✅ Even though you didn't complete all challenges I'm glad you tried the Unit Tests! I can't stress enough how important this is. As you said it yourself in the Readme, it helps you write better code when you do it with tests in mind! A great first step would be isolating validation into it's own static class.🤯 I'm also very happy with the way you handled updates, letting the user choose which property to update 👏👏

🥋You're only one project away from getting your Olive Green belt! 🫒🫒 The Habit Tracker 📊

🔍️The code is mostly ok, except for a couple of things. First you're using recursion (a method calling itself). This is ok for small projects but it can get messy and hard to debug really quickly, so I'd refrain from it (I didn't know better when making the tutorial)

🔍️Your project does have a configuration file but it's not a json. For future projects make sure you use a json since an xml file is an old practice that the .NET ecosystem has long moved away from.  .NET was redesigned for cloud, containers, and cross-platform apps, and JSON fits that world better than XML.

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 

Ledana:
Thank you my friend so much for this. i know that no matter how clean a code may get, it can actually be cleaner. im now getting used to leave main as plain and simple as possible, about static classes to isolate validation thank so much for the advice, ill try to follow it. the json file you are right i think, its just when i add a config file in the proj the xml is what vs opens for me. but ill see next time for json cause in my learning journy im understanding that saving a connectionstring in the right place helps a lot in future coding. again thank you, ill keep in mind your siggestions and advices. P.S you website is amazing thank you for the work!

PR #272, opened 2026-07-11
Project Finished..!

Sangeerths:


TheCSharpAcademy:
@Sangeerths **Project approved!** 😄✅

Excellent work, and well done for completing the optional challenges too! There is a lot to like here: strict date validation, prevention of negative manual sessions, Dapper-based data access, configuration through `appsettings.json`, good use of Spectre.Console, filtering, sorting, and a detailed README.

🥋 **Congratulations! You've earned your Olive Green Belt!** 🫒🫒

---

# 📋 Feedback

### 🔍 Stopwatch duration isn't being saved

Your stopwatch calculates a local `TimeSpan`:

```csharp
TimeSpan duration = session.EndTime - session.StartTime;
```

but never assigns it to:

```csharp
session.Duration;
```

As a result, the object passed to `InsertCodingSession` keeps the default `float` value of `0`.

---

### 🔍 Simplify `GetCodingSessions`

Your database method currently returns:

```csharp
IEnumerable<List<CodingSessionModel>>
```

which wraps a list inside another list. This forces you to use nested loops when displaying the sessions.

Dapper can return the list directly:

```csharp
public List<CodingSessionModel> GetCodingSessions()
{
    using var connection = new SqliteConnection(_connectionString);

    return connection
        .Query<CodingSessionModel>(
            "SELECT * FROM CodingSessions;")
        .ToList();
}
```

This makes the code simpler and easier to follow.

---

### 🔍 Improve the date input experience

If I enter an invalid **End Time**, I have to start over and re-enter the **Start Time** as well.

A better user experience would be to keep the valid start time in memory and only ask for the end time again until it passes validation.

---

### 🔍 Updating records

Currently users have to update every property.

It would provide a better user experience if they could choose which property they want to edit instead of re-entering the entire session.

---

### 🔍 Use `Microsoft.Data.Sqlite`

You're currently using the wrong SQLite library. Because of that you have to initialize the `Batteries` class.

Using `Microsoft.Data.Sqlite` removes that dependency and is the library we recommend throughout the curriculum.

image

---

### 🔍 Unit tests

Unit tests are optional, but this project would really benefit from a few.

Validation methods and duration calculation are excellent candidates, especially around the stopwatch logic and ensuring the end time cannot be earlier than the start time.

---

### 🕵️ Code conventions

Have a look through our style guide. I noticed a few convention issues—for example, many files contain unused `using` directives.

📖 **Code Conventions:**
https://www.thecsharpacademy.com/article/58/code-conventions

image

---

# 🗺️ Suggested next step

I noticed that the only course you've completed so far is **Introduction to Object-Oriented Programming**.

I'd highly recommend taking our **LINQ Fundamentals** course next. LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive.

https://thecsharpacademy.com/course/8/article/1/500249/False

---

# ☕

If you're enjoying our roadmap and would like to support the Academy, please consider buying us a coffee:

https://thecsharpacademy.com/Contribute

We really appreciate it! 🙂

---

# 👪 Join the community

If you haven't already, come and join our Discord community!

Don't forget to add your Discord username to your Academy profile so other students and reviewers can easily find you.

https://thecsharpacademy.com/dashboard/profile

---

Great work overall! ✋🏻

Looking forward to seeing your next projects! 🚀

PR #274, opened 2026-07-15
felikshetalia's contribution for CodingTracker

felikshetalia:


TheCSharpAcademy:
@felikshetalia **Project approved!** 😄✅ This is an excellent submission. You went far beyond meeting all of the project requirements, well done sir!

---

# 📋 Feedback

💖 The project is super well structured, it shows you're starting to think about architecture early, which is great. 

---

### 🔍 Inconsistent error handling

 You said it yourself in your Readme: although there's good error handling coverage it's a bit inconsistent. At this stage i'd let exceptions bubble up and get caught with a general exception. I'd only catch specific exceptions if I want to do handle it differently (i.e. logging db exceptions).

---

### 🔍 Input is cumbersome

Users need to type too much. There should be an option to let the users choose "today". Also, if end time input doesn't pass validation, the whole things starts again including start time. And when update you force users to update every property, defeating the purpose of an update function (what's the difference between add and update if I need to update every property?)

---

### 🔍 Mapping is repetitive.

You've also mentioned this yourself, but for repeated mapping it's better to create an extension method. Something like:
```
private static CodingSession? ToCodingSession(CodingSessionDTO dto)
```

---

### 🔍 Updating records

Currently users have to update every property.

It would provide a better user experience if they could choose which property they want to edit instead of re-entering the entire session.

---

# 🗺️ Suggested next step

I noticed that you've completed **Introduction to Object-Oriented Programming** **and intro to SQL**.

I'd highly recommend taking our [**LINQ Fundamentals** ](https://thecsharpacademy.com/course/8/article/1/500249/False)course next. LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive.

---

# ☕

If you're enjoying our roadmap and would like to support the Academy, please consider buying us a coffee:

https://thecsharpacademy.com/Contribute

We really appreciate it! 🙂

---

Great work overall! ✋🏻

Looking forward to seeing your next projects! 🚀

PR #276, opened 2026-07-22
Ready for review

Solomonlol:


TheCSharpAcademy:
@Solomonlol Project approved! 😄✅ And well done for completing the stopwatch challenge. 

🥋You still need to submit your [Freecodecamp certificate](https://thecsharpacademy.com/project/75/freecodecamp-certification) before you start earning belts 😊

---

## **Feedback**

🔍️ When updating record it would be good to let the users choose with property to update. That defeats the purpose of the update, since the effort for the user is the same as adding a new record. Additionally if I have a validation error I get sent back to the menu. Ideally want to let me try again in that same operation.

---

🔍️ Currently you store, Date, StartTime, EndTime as three separate strings.

Although SQLite stores dates as text anyway (as noted in the project tips), you could store a full ISO-8601 timestamp (for example "2026-07-29T14:30:00") for start and end times. That simplifies sorting, filtering, and future reporting while still meeting SQLite's text-storage approach.

---

🔍️ This is probably the biggest architectural improvement I'd suggest. CodingController currently prints tables, validates input, reads keyboard, talks to SQLite, creates models, starts timer. That's a lot.

As projects grow, splitting those concerns into dedicated classes makes everything easier to test and maintain.

---

🗺️ I noticed you've completed a few courses in the academy already. Way to go! Next I'd highly recommend [taking our **LINQ Fundamentals** course next](https://www.thecsharpacademy.com). LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive. 

---

☕ If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

---

Overall great work! ✋🏻 Looking forward to seeing your next projects!

PR #280, opened 2026-08-09
Project and extra tasks are done!

eltab0:


TheCSharpAcademy:
@eltab0   Project approved! 😄✅ This is a strong Coding Tracker submission. The required architecture and functionality are there, and several optional challenges have also been completed. The project shows good separation of concerns, sensible validation, Dapper-based data access, and you even included a test project, well done!!

👩‍🏫Feedback
---

🔍️ If the SQLite database is corrupted, the current delete/update code can still crash because calls such as connection.Open() or connection.Execute(...) may throw a SqliteException. Checking whether an ID exists does not protect against that kind of failure.

A better pattern would be to catch database exceptions around the actual data-access operation and show a user-friendly message.

---

🔍️There is a nullable-reference-type issue here:
```
internal static bool IsSessionMissing(CodingSession codingSession, int id)
```
but the method explicitly checks:
```
if (codingSession is null)
```
So the parameter should be:
```
CodingSession? codingSession
```
---

🔍️ Your Readme is really good and detailed. Well done for that. It does feel it's been written by AI and I can't really "hear your voice". I'd love to see you write your own words in the next Readme. 😊

🧐Challenges Review
---

✅ Stopwatch functionality completed. CodingStopwatch records DateTime.Now at the beginning and end of a live coding session, displays elapsed time, and allows the result to be saved. This is a very nice implementation of the stopwatch challenge.

✅ Filtering and sorting completed. Users can filter by the last day, week, month, year, or a custom range, and can choose oldest-first or latest-first ordering. FilterQueryBuilder() also builds the query dynamically while keeping values parameterized.

✅ Unit tests completed. There are tests for exact date parsing, start/end ordering, and numeric validation. These are appropriate targets because they are pure logic and do not depend on the database.

The tests could be expanded slightly with edge cases such as equal start/end values, leap-year dates, malformed whitespace, and an invalid custom range where from > to.

---
🗺️ **Suggested next step:** 
I noticed that the only course you've completed so far is [**Introduction to Object-Oriented Programming**](
https://thecsharpacademy.com/course/8/article/1/500249/False). I'd highly recommend taking our **LINQ Fundamentals** course next. LINQ is one of the most important tools in modern C# and becoming comfortable with it will make your code much cleaner and more expressive.

---

👪Oh and if possible join our Discord Community and add your username [to your profile](https://thecsharpacademy.com/dashboard/profile) so you can be part of the community as well. 😁 And do you mind adding a country in your profile as well? The dashboard and leaderboard look neater when there's a flag. If you're not comfortable with that you can submit a random country.

---

☕If you like our roadmap, please consider [buying us a coffee](https://thecsharpacademy.com/Contribute). We appreciate your help 🙂

Overall great work!✋🏻Looking forward to seeing your next projects!

